PR #18, opened 2023-08-21
Enable specifying api key in config

joewagner:
This enables specifying an api key that bypasses the Validator's rate limiting.
You can pass an `apiKey` parameter to the `Validator` and `Database` classes, which is converted to the correct header and send with read requests.

Example use:
```
new Validator({ apiKey, ... });
new Database({ apiKey, ... });
```

### DIFF
--- a/packages/local/package.json
+++ b/packages/local/package.json
@@ -45,7 +45,7 @@
   "scripts": {
     "postinstall": "cd registry && npm install",
     "test": "mocha",
-    "test:ci": "npm run build && mocha --exit",
+    "test:ci": "mocha --exit",
     "lint": "eslint '**/*.{js,ts}'",
     "lint:fix": "npm run lint -- --fix",
     "prepublishOnly": "npm run build",
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -59,8 +59,8 @@
   },
   "scripts": {
     "prepublishOnly": "npm run build",
-    "test": "npm run test:browser && npm run test:dev",
-    "test:dev": "mocha --exit",
+    "test": "npm run test:browser && npm run test:mocha",
+    "test:mocha": "mocha --exit",
     "test:browser": "npx playwright install --with-deps && PW_TS_ESM_ON=true playwright test",
     "test:browser-show-report": "npx playwright show-report",
     "test:ci": "npm run coverage",
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -59,7 +59,8 @@
   },
   "scripts": {
     "prepublishOnly": "npm run build",
-    "test": "npm run test:browser && mocha --exit",
+    "test": "npm run test:browser && npm run test:dev",
+    "test:dev": "mocha --exit",
     "test:browser": "npx playwright install --with-deps && PW_TS_ESM_ON=true playwright test",
     "test:browser-show-report": "npx playwright show-report",
     "test:ci": "npm run coverage",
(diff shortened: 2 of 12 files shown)


packages/local/package.json
```diff
@@ -45,7 +45,7 @@
   "scripts": {
     "postinstall": "cd registry && npm install",
     "test": "mocha",
-    "test:ci": "npm run build && mocha --exit",
```
joewagner, line 48: Not related to this PR, but lerna and nx handle the need to build (or not) before running the test.  Having this written into the test command was causing issues.

packages/sdk/package.json
```diff
@@ -59,7 +59,8 @@
   },
   "scripts": {
     "prepublishOnly": "npm run build",
-    "test": "npm run test:browser && mocha --exit",
+    "test": "npm run test:browser && npm run test:mocha",
+    "test:mocha": "mocha --exit",
```
joewagner, line 63: I added this command to make it easier for devs to just run the mocha tests since the browser tests take so long.

packages/sdk/src/helpers/config.ts
```diff
@@ -1,11 +1,13 @@
 import fs from "node:fs";
 import { type WaitableTransactionReceipt } from "../registry/utils.js";
+import { type FetchConfig } from "../validator/client/index.js";
 import { type ChainName, getBaseUrl } from "./chains.js";
 import { type Signer, type ExternalProvider, getSigner } from "./ethers.js";
 
 export interface ReadConfig {
   baseUrl: string;
   aliases?: AliasesNameMap;
+  apiKey?: string;
```
joewagner, line 10: The usage is now:
```
new Validator({ apiKey, ... });
new Database({ apiKey, ... });
```

packages/sdk/src/helpers/config.ts
```diff
@@ -98,3 +100,16 @@ export function jsonFileAliases(filepath: string): AliasesNameMap {
     },
   };
 }
+
+export function prepReadConfig(config: Partial<ReadConfig>): FetchConfig {
+  const conf: FetchConfig = {};
+  if (config.apiKey) {
+    conf.init = {
+      headers: {
+        "Api-Key": config.apiKey,
```
joewagner, line 109: This is the bulk of the logic.  Take the `apiKey` param and set the `Api-Key` request header with the given value.

packages/sdk/src/validator/query.ts
```diff
@@ -1,4 +1,4 @@
-import { type Signal } from "../helpers/await.js";
+import { type Signal } from "../helpers/index.js";
```
joewagner, line 1: unrelated cleanup

packages/sdk/test/validator/config.json
```diff
@@ -3,7 +3,8 @@
   "HTTP": {
     "Port": "8081",
     "RateLimInterval": "1s",
-    "MaxRequestPerInterval": 10
+    "MaxRequestPerInterval": 10,
+    "APIKey": "foo"
```
joewagner, line 7: Keeping the tests simple for the time being.  The api key is always going to be "foo"

As an aside in regard to the api key implementation in the validator... Since the key is a fixed config value we should probably ensure the value is only ever used on the server side of the Studio.
Otherwise anyone who logs into the Studio will be able to get the api key and use it as much as they like.

carsonfarmer:
Is this still WIP @joewagner?

joewagner:
> Is this still WIP @joewagner?

It's ready for review, but it's aiming at the cleanup branch, which I'll merge and then point this at main.

packages/sdk/test/validator/config.json
asutula: Vercel makes it hard to expose secrets to the client. By default all env vars are server side only. You have to opt in to exposing them to the client by prefixing the env var name with `NEXT_PUBLIC_`.

asutula (approved):
This looks great. Nice work. I'm fine with this as is, but just want to surface an idea... What about making the config option called `extraHeaders` and it's type be `{ [key: string]: string }`. I'm just thinking this is, at this point at least, kind of a bandaid fix for Studio and not a real feature we want to advertise. If we leave the config option as kinda generic like that, we don't have to explain much about it in the docs.

PR #21, opened 2023-08-23
chore: update versions for cli and sdk

joewagner:
As part of publishing on yesterday, the versions of cli and sdk have been updated.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12652,7 +12652,7 @@
     },
     "packages/cli": {
       "name": "@tableland/cli",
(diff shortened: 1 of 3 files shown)


carsonfarmer:
In the future, this will "just work" with lerna, right?

carsonfarmer (approved):

joewagner:
> In the future, this will "just work" with lerna, right?

Yeah, with the understanding that when you make a PR that includes a publishable change, you should be including the version change with the PR.

PR #22, opened 2023-08-24
throw an error if the registry contract hasn't been deployed

joewagner:
## Overview
As an improvement working toward fixing https://github.com/tablelandnetwork/local-tableland/issues/379 this PR introduces behavior that check if the registry has been deployed before starting the Validator.   If the contract is **not** deployed an error is thrown.  This will make CI tests fail immediately and obviously, which will make the reason for failure more obvious.

## Details
Along with requiring the existence of the registry on chain for successful startup, there are a few other changes:
- This reverts some of the port failure retry logic since that didn't solve the issue with CI failures.
- This moves the startup and shutdown tests into their own test file
- This moves the hard coded registry contract address up one level, from the validator.ts file to the main.ts file. I needed to do this to enable checking the existence of the deployed contract.  This also moves one step closer to getting the contract out of the deploy, which isn't possible atm since the deploy fails silently.

NOTE: this is based off the `publish` branch.  Once that's merged this will be pointed at main.

### DIFF
--- a/packages/local/src/main.ts
+++ b/packages/local/src/main.ts
@@ -154,6 +154,8 @@ class LocalTableland {
     // wait until initialization is done
     await waitForReady(registryReadyEvent, this.initEmitter);
 
+    await new Promise((resolve) => setTimeout(resolve, 5000));
+
     this._deployRegistry();
 
     const deployed = await this.#_ensureRegistry();
--- a/packages/local/src/main.ts
+++ b/packages/local/src/main.ts
@@ -5,11 +5,13 @@ import { type ChildProcess } from "node:child_process";
 import { EventEmitter } from "node:events";
 import spawn from "cross-spawn";
 import shell from "shelljs";
+import { getDefaultProvider } from "ethers";
 import { chalk } from "./chalk.js";
 import { ValidatorDev, ValidatorPkg } from "./validators.js";
 import {
   buildConfig,
   type Config,
+  checkPortInUse,
   defaultRegistryDir,
   inDebugMode,
   isValidPort,
@@ -22,11 +24,11 @@ import {
   getValidator,
   logSync,
   pipeNamedSubprocess,
-  probePortInUse,
   waitForReady,
 } from "./util.js";
 
 const spawnSync = spawn.sync;
+const registryAddress = "0xe7f1725e7734ce288f8367e1bb143e90bb3f0512";
 
 class LocalTableland {
   config;
@@ -62,7 +64,7 @@ class LocalTableland {
       typeof config.registryDir === "string" &&
       config.registryDir.trim() !== ""
     ) {
-      this.registryDir = config.registryDir;
+      this.registryDir = config.registryDir.trim();
     } else {
       this.registryDir = defaultRegistryDir();
     }
@@ -80,19 +82,15 @@ class LocalTableland {
   }
 
   async #_start(config: Config = {}): Promise<void> {
-    if (
-      typeof this.registryDir !== "string" ||
-      this.registryDir.trim() === ""
-    ) {
+    if (typeof this.registryDir !== "string" || this.registryDir === "") {
       throw new Error("cannot start a local network without Registry");
     }
 
     // make sure we are starting fresh
-    // TODO: I don't think this is doing anything anymore...
     this.#_cleanup();
 
     // Check if the hardhat port is in use (defaults to 5 retries, 300ms b/w each try)
-    const registryPortIsTaken = await probePortInUse(this.registryPort);
+    const registryPortIsTaken = await checkPortInUse(this.registryPort);
     // Note: this generally works, but there is a chance that the port will be
     // taken but returns `false`. E.g., try racing two instances at *exactly*
     // the same, and `EADDRINUSE` occurs. But generally, it'll work as expected.
@@ -101,14 +99,18 @@ class LocalTableland {
     // Else, notify the user only if it's a not the default and is custom.
     if (registryPortIsTaken) {
       throw new Error(`port ${this.registryPort} already in use`);
-    } else {
-      // Notify that we're using a custom port since it's not the default 8545
+    }
+
+    // Notify that we're using a custom port since it's not the default 8545
+    if (
       this.registryPort !== this.defaultRegistryPort &&
-        shell.echo(
-          `[${chalk.magenta.bold("Notice")}] Registry is using custom port ${
-            this.registryPort
-          }`
-        );
+      this.silent !== true
+    ) {
+      shell.echo(
+        `[${chalk.magenta.bold("Notice")}] Registry is using custom port ${
+          this.registryPort
+        }`
+      );
     }
 
     // You *must* store these in `process.env` to access within the hardhat subprocess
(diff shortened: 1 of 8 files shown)


packages/local/src/main.ts
```diff
@@ -22,12 +24,15 @@ import {
...
   pipeNamedSubprocess,
-  probePortInUse,
   waitForReady,
 } from "./util.js";
 
 const spawnSync = spawn.sync;
 
+// TODO: maybe this can be parsed out of the deploy process?
+// Since it's always the same address just hardcoding for now.
+const registryAddress = "0xe7f1725e7734ce288f8367e1bb143e90bb3f0512";
```
joewagner, line 34: This used to reside in the validator.ts file, but we need access to it here now so it's been moved.  If we can find a way to get this value from the deploy process, that would be ideal.  Until then this works.

packages/local/src/main.ts
```diff
@@ -62,7 +67,7 @@ class LocalTableland {
       typeof config.registryDir === "string" &&
       config.registryDir.trim() !== ""
     ) {
-      this.registryDir = config.registryDir;
+      this.registryDir = config.registryDir.trim();
```
joewagner, line 70: trimming before setting the instance value, which means all later access doesn't need to trim

packages/local/src/main.ts
```diff
@@ -80,19 +85,15 @@ class LocalTableland {
   }
 
   async #_start(config: Config = {}): Promise<void> {
-    if (
-      typeof this.registryDir !== "string" ||
-      this.registryDir.trim() === ""
```
joewagner, line 85: See above...  This has already been trimmed.

packages/local/src/main.ts
```diff
@@ -80,19 +85,15 @@ class LocalTableland {
...
+    if (typeof this.registryDir !== "string" || this.registryDir === "") {
       throw new Error("cannot start a local network without Registry");
     }
 
     // make sure we are starting fresh
-    // TODO: I don't think this is doing anything anymore...
     this.#_cleanup();
 
     // Check if the hardhat port is in use (defaults to 5 retries, 300ms b/w each try)
-    const registryPortIsTaken = await probePortInUse(this.registryPort);
```
joewagner, line 95: An old PR attempted to fix the CI issues by retrying the same hardhat node port in hopes it will become available.  The issue comes from a crashing contract deploy process, **not** a failure to start the hardhat node.  We no longer need this retry logic. I've left the logic that ensures the port is available, since this is a nice mechanism to fail  fast and loud if the port was already taken.

packages/local/src/main.ts
```diff
@@ -101,14 +102,18 @@ class LocalTableland {
     // Else, notify the user only if it's a not the default and is custom.
     if (registryPortIsTaken) {
       throw new Error(`port ${this.registryPort} already in use`);
-    } else {
```
joewagner, line 104: Some cleanup here.  We don't need an else clause because throwing breaks out of this function.

packages/local/src/main.ts
```diff
@@ -101,14 +102,18 @@ class LocalTableland {
...
+
+    // Notify that we're using a custom port since it's not the default 8545
+    if (
       this.registryPort !== this.defaultRegistryPort &&
-        shell.echo(
-          `[${chalk.magenta.bold("Notice")}] Registry is using custom port ${
-            this.registryPort
-          }`
-        );
+      this.silent !== true
```
joewagner, line 110: This was getting logged even if the dev had set `silent: true`

packages/local/src/main.ts
```diff
@@ -149,17 +154,14 @@ class LocalTableland {
...
-          cwd: this.registryDir,
-        }
-      ),
-      !inDebugMode()
-    );
+    this._deployRegistry();
+
+    const deployed = await this.#_ensureRegistry();
+    if (!deployed) {
+      throw new Error(
```
joewagner, line 161: This ensures we fail fast and loud if the deploy failed.  At the moment you have to look through the CI logs and see if the failing tests are becaseue the registry wasn't deployed, or for some legitimate reason.

packages/local/src/main.ts
```diff
@@ -220,6 +222,33 @@ class LocalTableland {
...
+  }
+
+  async #_ensureRegistry(): Promise<boolean> {
+    const provider = getDefaultProvider(
+      `http://127.0.0.1:${this.registryPort}`
+    );
+    const code = await provider.getCode(registryAddress);
+
+    // if the contract exists, and is not empty, code will not be equal to 0x
+    return code !== "0x";
```
joewagner, line 249: To check if the registry was deployed, we can use the ethers.js `getCode` method.  If the contract exists it will have a long string corresponding to the contract code.  If not it will equal "0x".

packages/local/test/e2e.test.ts
```diff
@@ -1,220 +1,17 @@
-import { type Server } from "node:net";
```
joewagner, line 1: The changes in this file are the result of moving the network startup and shutdown tests to their own file.

packages/local/test/start.test.ts
```diff
@@ -0,0 +1,196 @@
...
+      (async function () {
+        await lt.start();
+      })()
+    ).to.be.rejectedWith(`port ${defaultPort} already in use`);
+  });
+
+  test("fails to start due to registry deploy failure", async function () {
+    const lt = new LocalTableland({ silent: true });
+    // note: need to cast as `any` since the method is private
+    const deployStub = stub(lt, "_deployRegistry" as any);
```
joewagner, line 90: I created a noop stub so the registry isn't deployed.  This must result in an error instead of successful startup.

All the other tests in this file were moved from the e2e file.

packages/local/src/main.ts
carsonfarmer: We can't just get this from the underlying client package?

packages/local/src/main.ts
```diff
@@ -220,6 +222,33 @@ class LocalTableland {
     console.log("\n\n*************************************\n");
   }
 
+  // note: Tests are using sinon to stub this method. Because typescript compiles ecmascript
+  //       private features, i.e. hash syntax, in a way that does not work with sinon we must
```
carsonfarmer, line 226: Wait, really? Huh, there are apparently other libs that have a problem with this. Do you have a reference for the issue? Just for interest sake?

packages/local/src/main.ts
carsonfarmer: Smart

carsonfarmer (approved):
Really nice work here @joewagner!

packages/local/src/main.ts
joewagner: by the client you mean `@tableland/evm`?

packages/local/src/main.ts
joewagner: Great question, I couldn't find an issue in Sinon.  Maybe I should open one?  Here's the Sinon source https://github.com/sinonjs/sinon/blob/main/lib/sinon/util/core/is-non-existent-property.js#L10
When you have a class the uses the js [private class feature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields), you'll always get `false` when you do something like:
``` 
class Derp {
  #_foo() {
    return "foo";
  }
}

const instance = new Derp();

console.log("#_foo" in instance); // => false
console.log(instance.#_foo); // => type error
```

Maybe it's impossible to spy on private fields?

joewagner:
@carsonfarmer This was automatically changed to point at main after merging #21, can you give another ✅ ?

joewagner:
@asutula I added the fix for the SDK regression here since without the changes in this PR the CI fails constantly.
Can you give a 👍 to this?

asutula (approved):

PR #36, opened 2023-09-24
fix abort controller handling

joewagner:
The initial work for this PR is located in https://github.com/tablelandnetwork/js-tableland/pull/576/files

_Summary
Fixes abort controller handling which allows a user to abort and configure the polling timeout and interval for SDK methods that poll the gateway for a receipt._

_Details
The core of this PR was adding the wiring to make the underlying polling handling aware of the user's abort controller. However, I found the API to be unintuitive and spent some time refactoring how a user goes about providing a abort-able controller with a custom timeout and polling interval. See comments below for details._

_NOTE: This would be a breaking API change and therefore would require a major version bump. We probably could make the core fix while not breaking the API, but I don't think there's any need to be bashful about frequently releasing new major versions at this point. What do you guys think?_

### DIFF
--- a/packages/sdk/src/database.ts
+++ b/packages/sdk/src/database.ts
@@ -45,7 +45,7 @@ export class Database<D = unknown> {
    */
   static readOnly(chainNameOrId: ChainName | number): Database {
     console.warn(
-      "`Database.readOnly()` is a depricated method, use `new Database()`"
+      "`Database.readOnly()` is a deprecated method, use `new Database()`"
     );
     const baseUrl = getBaseUrl(chainNameOrId);
     return new Database({ baseUrl });
@@ -87,7 +87,7 @@ export class Database<D = unknown> {
    * @returns An array of run results.
    */
   //    Note: if we want this package to mirror the D1 package in a way that
-  //    enables compatability with packages built to exend D1, then the return type
+  //    enables compatability with packages built to extend D1, then the return type
   //    here will potentially affect if/how those packages work.
   //    D1-ORM is a good example: https://github.com/Interactions-as-a-Service/d1-orm/
   async batch<T = D>(
@@ -124,9 +124,7 @@ export class Database<D = unknown> {
       // and return an Array of the query results.
       if (type === "read") {
         return await Promise.all(
-          statements.map(
-            async (stmt) => await stmt.all<T>(undefined, controller)
-          )
+          statements.map(async (stmt) => await stmt.all<T>({ controller }))
         );
       }
 
@@ -202,7 +200,7 @@ export class Database<D = unknown> {
       const { statements } = await normalize(statementStrings);
       const count = statements.length;
       const statement = this.prepare(statementStrings);
-      const result = await statement.run(controller);
+      const result = await statement.run({ controller });
       // Adds a count property which isn't typed
       result.meta.count = count;
       return result;
@@ -252,7 +250,7 @@ async function normalizedToRunnables(
         // check if these tables are in the normalized table names
         // if so, filter them out (i.e., they are not being mutated)
         const filteredTables = norm.tables.filter(
-          (tableName) => !tableNames.includes(tableName)
+          (tableName: string) => !tableNames.includes(tableName)
         );
         // if the filtered tables are greater than 1, then there are two
         // tables being mutated in a single statement, which is not allowed
--- a/packages/sdk/src/database.ts
+++ b/packages/sdk/src/database.ts
@@ -4,12 +4,12 @@ import { wrapResult } from "./registry/utils.js";
 import {
   type Config,
   type AutoWaitConfig,
+  type ChainName,
+  type PollingController,
+  type Signer,
   checkWait,
   extractBaseUrl,
-  type ChainName,
   getBaseUrl,
-  type Signal,
-  type Signer,
   normalize,
   validateTableName,
 } from "./helpers/index.js";
@@ -83,7 +83,7 @@ export class Database<D = unknown> {
    * in the sequence fails, then an error is returned for that specific
    * statement, and it aborts or rolls back the entire sequence.
    * @param statements A set of Statement objects to batch and submit.
-   * @param opts Additional options to control execution.
+   * @param controller An optional object used to control receipt polling behavior.
    * @returns An array of run results.
    */
   //    Note: if we want this package to mirror the D1 package in a way that
@@ -92,8 +92,8 @@ export class Database<D = unknown> {
   //    D1-ORM is a good example: https://github.com/Interactions-as-a-Service/d1-orm/
   async batch<T = D>(
     statements: Statement[],
-    opts: Signal = {}
-    // reads returns an Array with legnth equal to the number of batched statements,
+    controller?: PollingController
+    // reads returns an Array with length equal to the number of batched statements,
     // everything else a single result wrapped in an Array for backward compatability.
   ): Promise<Array<Result<T>>> {
     try {
@@ -124,7 +124,9 @@ export class Database<D = unknown> {
       // and return an Array of the query results.
       if (type === "read") {
         return await Promise.all(
-          statements.map(async (stmt) => await stmt.all<T>(undefined, opts))
+          statements.map(
+            async (stmt) => await stmt.all<T>(undefined, controller)
+          )
         );
       }
 
@@ -135,7 +137,8 @@ export class Database<D = unknown> {
           await execCreateMany(
             this.config,
             statements.map((stmt) => stmt.toString())
-          )
+          ),
+          controller
         );
 
         // TODO: wrapping in an Array is required for back compat, consider changing this for next major
@@ -160,7 +163,8 @@ export class Database<D = unknown> {
 
       const receipt = await checkWait(
         this.config,
-        await execMutateMany(this.config, runnables)
+        await execMutateMany(this.config, runnables),
+        controller
       );
 
       // TODO: wrapping in an Array is required for back compat, consider changing this for next major
@@ -186,19 +190,19 @@ export class Database<D = unknown> {
    * transaction. In the future, more "intelligent" transaction planning,
    * splitting, and batching may be used.
    * @param statementStrings A set of SQL statement strings separated by semi-colons.
-   * @param opts Additional options to control execution.
+   * @param controller An optional object used to control receipt polling behavior.
    * @returns A single run result.
    */
   async exec<T = D>(
     statementStrings: string,
-    opts: Signal = {}
+    controller?: PollingController
   ): Promise<Result<T>> {
     // TODO: Note that this method appears to be the wrong return type in practice.
     try {
       const { statements } = await normalize(statementStrings);
       const count = statements.length;
       const statement = this.prepare(statementStrings);
-      const result = await statement.run(opts);
+      const result = await statement.run(controller);
       // Adds a count property which isn't typed
       result.meta.count = count;
       return result;
@@ -213,9 +217,9 @@ export class Database<D = unknown> {
   /**
    * Export a (set of) tables to the SQLite binary format.
    * Not implemented yet!
-   * @param _opts Additional options to control execution.
+   * @param controller An optional object used to control receipt polling behavior.
    */
-  async dump(_opts: Signal = {}): Promise<ArrayBuffer> {
+  async dump(_controller?: PollingController): Promise<ArrayBuffer> {
     throw errorWithCause("DUMP_ERROR", new Error("not implemented yet"));
   }
 }
--- a/packages/sdk/src/helpers/await.ts
+++ b/packages/sdk/src/helpers/await.ts
@@ -1,17 +1,19 @@
 export type Awaitable<T> = T | PromiseLike<T>;
 
 export interface Signal {
-  signal?: AbortSignal;
+  signal: AbortSignal;
+  abort: () => void;
 }
 
 export interface Interval {
-  interval?: number;
+  interval: number;
+  cancel: () => void;
 }
 
-export type SignalAndInterval = Signal & Interval;
+export type PollingController = Signal & Interval;
 
 export interface Wait<T = unknown> {
-  wait: (opts?: SignalAndInterval) => Promise<T>;
+  wait: (controller?: PollingController) => Promise<T>;
 }
 
 export interface AsyncData<T> {
@@ -21,36 +23,41 @@ export interface AsyncData<T> {
 
 export type AsyncFunction<T> = () => Awaitable<AsyncData<T>>;
 
-export function getAbortSignal(
-  signal?: AbortSignal,
-  maxTimeout: number = 60_000
-): {
-  signal: AbortSignal;
-  timeoutId: ReturnType<typeof setTimeout> | undefined;
-} {
-  let abortSignal: AbortSignal;
-  let timeoutId;
-  if (signal == null) {
-    const controller = new AbortController();
-    abortSignal = controller.signal;
-    // return the timeoutId so the caller can cleanup
-    timeoutId = setTimeout(function () {
+export function createSignal(): Signal {
+  const controller = new AbortController();
+  return {
+    signal: controller.signal,
+    abort: () => {
       controller.abort();
-    }, maxTimeout);
-  } else {
-    abortSignal = signal;
-  }
-  return { signal: abortSignal, timeoutId };
+    },
+  };
+}
+
+export function createPollingController(
+  timeout: number = 60_000,
+  pollingInterval: number = 1500
+): PollingController {
+  const controller = new AbortController();
+  const timeoutId = setTimeout(function () {
+    controller.abort();
+  }, timeout);
+  return {
+    signal: controller.signal,
+    abort: () => {
+      controller.abort();
+    },
+    interval: pollingInterval,
+    cancel: () => {
+      clearTimeout(timeoutId);
+    },
+  };
 }
 
 export async function getAsyncPoller<T = unknown>(
   fn: AsyncFunction<T>,
-  interval: number = 1500,
-  signal?: AbortSignal
+  controller?: PollingController
 ): Promise<T> {
-  // in order to set a timeout other than 10 seconds you need to
-  // create and pass in an abort signal with a different timeout
-  const { signal: abortSignal, timeoutId } = getAbortSignal(signal, 10_000);
+  const control = controller ?? createPollingController();
   const checkCondition = (
     resolve: (value: T) => void,
     reject: (reason?: any) => void
@@ -59,15 +66,15 @@ export async function getAsyncPoller<T = unknown>(
       .then((result: AsyncData<T>) => {
         if (result.done && result.data != null) {
           // We don't want to call `AbortController.abort()` if the call succeeded
-          clearTimeout(timeoutId);
+          control.cancel();
           return resolve(result.data);
         }
-        if (abortSignal.aborted) {
+        if (control.signal.aborted) {
           // We don't want to call `AbortController.abort()` if the call is already aborted
-          clearTimeout(timeoutId);
-          return reject(abortSignal.reason);
+          control.cancel();
+          return reject(control.signal.reason);
         } else {
-          setTimeout(checkCondition, interval, resolve, reject);
+          setTimeout(checkCondition, control.interval, resolve, reject);
         }
       })
       .catch((err) => {
--- a/packages/sdk/src/helpers/await.ts
+++ b/packages/sdk/src/helpers/await.ts
@@ -1,32 +1,68 @@
+/**
+ * A type that can be awaited.
+ * @property T The type to await.
+ */
 export type Awaitable<T> = T | PromiseLike<T>;
 
+/**
+ * A signal to abort a request.
+ * @property signal The {@link AbortSignal} to abort a request.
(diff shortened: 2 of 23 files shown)


sanderpick:
Thanks for sorting this out @joewagner. Should we get this out soonish? Sounded like the timeout issue was blocking some folks on Filecoin.

joewagner:
> Thanks for sorting this out @joewagner. Should we get this out soonish? Sounded like the timeout issue was blocking some folks on Filecoin.

Sounds good to me!  I'm planning to make sure the studio users get support, but otherwise can work on this.

sanderpick:
@joewagner I asked @dtbuchholz if he could take this over... sounds like the abort bug is the main cause of pain for the hackathon so figured we should try to get this out.

dtbuchholz:
i tested these changes in the Studio with just a symlink (no code changes to the Studio...and just using the default controller functionality in this PR). occasionally, it works on sepolia, but it's still aborting. in this video, you'll notice:
1. the tx is signed, the SDK starts waiting for confirmation, and it gets confirmed onchain ~20 seconds later—the tx is logged in the console.
2. thereafter, **it takes 13 seconds** for the validator to materialize the table. it makes 8 requests without success and then aborts.

it seems like the validator is having issues materializing sepolia data?

https://github.com/tablelandnetwork/tableland-js/assets/13358940/46643199-1ab3-49b7-aaa2-3785d410f4d6

packages/sdk/src/database.ts
```diff
@@ -124,7 +124,9 @@ export class Database<D = unknown> {
       // and return an Array of the query results.
       if (type === "read") {
         return await Promise.all(
-          statements.map(async (stmt) => await stmt.all<T>(undefined, opts))
+          statements.map(
+            async (stmt) => await stmt.all<T>(undefined, controller)
+          )
```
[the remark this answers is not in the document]
sanderpick: cool, trying to refresh my memory here. shouldn't your array have 5 entries?

sanderpick:
> * regarding [this comment](https://github.com/tablelandnetwork/js-tableland/pull/576#issuecomment-1660691750), is it something worthwhile implementing, or should we keep this as-is?

hmm, i'm not totally clear on what needs to conform to the D1 API, and if we're planning on maintaining compatibility. my pref would be to make an options object like in my [example commit](https://github.com/tablelandnetwork/js-tableland/commit/ef8f74f7acb51709d02f1c3dada2f69a21b55dd1). 

> * in order to set a predefined per-chain polling timeout value, would it make sense to put that in evm-tableland's [network.ts file](https://github.com/tablelandnetwork/evm-tableland/blob/main/network.ts)? otherwise, it'd require custom chain handling in the [chains](https://github.com/tablelandnetwork/tableland-js/blob/main/packages/sdk/src/helpers/chains.ts) helper, which tries to avoid that by importing from evm-tableland. for example, network.ts could export a "chainPollingTimeout" or "chainBlockTime"...something like that.

Nice, I like the idea of using evm-tableland here. That could be integrated here in a diff PR.

sanderpick:
@dtbuchholz regarding the video, if you increase the timeout to like 60 seconds, it should work then right? ignoring the fact that it shouldn't be that slow on sepolia for now (@avichalp / @brunocalza any idea what's going on there? 13 seconds sounds pretty slow.)

joewagner:
> > * regarding [this comment](https://github.com/tablelandnetwork/js-tableland/pull/576#issuecomment-1660691750), is it something worthwhile implementing, or should we keep this as-is?
> 
> hmm, i'm not totally clear on what needs to conform to the D1 API, and if we're planning on maintaining compatibility. my pref would be to make an options object like in my [example commit](https://github.com/tablelandnetwork/js-tableland/commit/ef8f74f7acb51709d02f1c3dada2f69a21b55dd1). @joewagner do you have any insights here?
> 

Based on the [abstract class](https://github.com/cloudflare/workerd/blob/main/types/defines/d1.d.ts#L20) D1 publishes, it looks like we need to make the `first` method polymorphic and potentially accept a colName string parameter, but it can also be called with no arguments.  
This means we should be able to accept an optional `Options` that contains whatever properties we want to support in all of the statement methods.
maybe something like:
```
{
  first<T = unknown>(colName: string, opts?: TablelandOptions): Promise<T | null>;
  first<T = Record<string, unknown>>(opts?: TablelandOptions): Promise<T | null>;
  run<T = Record<string, unknown>>(opts?: TablelandOptions): Promise<D1Result<T>>;
  all<T = Record<string, unknown>>(opts?: TablelandOptions): Promise<D1Result<T>>;
  raw<T = unknown[]>(opts?: TablelandOptions): Promise<T[]>;
}
```

Any kind of third party tools that expect the SDK to be D1 compliant will be able to call those methods as ewpected, and we can still allow folks to supply an options object that indicates the timeout they want.

NOTE: looks like we currently do `(opts: SignalAndInterval = {})` instead of using the `?`, i.e. `(opts?: SignalAndInterval)`.  Our current way of setting a default feels nicer and has the same affact imo, but either should work.

joewagner:
I have a thought after looking at this with fresh eyes. Regardless if we go with per chain default timeouts, it would be nice to be able to set a global timeout via an [`overrideDefaults`](https://github.com/tablelandnetwork/tableland-js/pull/36) type of interface. 
This way, as a dev, I can decide how often my default polling should happen and how long to wait before stopping the polling.

brunocalza:
> @dtbuchholz regarding the video, if you increase the timeout to like 60 seconds, it should work then right? ignoring the fact that it shouldn't be that slow on sepolia for now (@avichalp / @brunocalza any idea what's going on there? 13 seconds sounds pretty slow.)

According to the [current config ](https://github.com/tablelandnetwork/go-tableland/blob/main/docker/deployed/testnet/api/config.json#L84), we're waiting for at least one block to consider the tx final. and the ethereum block time is 12s, so I guess that's kind of expected. Were we experiencing faster times before?

dtbuchholz:
> According to the [current config ](https://github.com/tablelandnetwork/go-tableland/blob/main/docker/deployed/testnet/api/config.json#L84), we're waiting for at least one block to consider the tx final. and the ethereum block time is 12s, so I guess that's kind of expected. Were we experiencing faster times before?

@brunocalza ah okay that makes sense. out of curiosity, what's the a reason we impose the 1 block depth for sepolia and mumbai when arb/op goerli use `0`?

sanderpick:
> Any kind of third party tools that expect the SDK to be D1 compliant will be able to call those methods as ewpected, and we can still allow folks to supply an options object that indicates the timeout they want.

@joewagner sounds good. @dtbuchholz do you want to give this a shot? 

> NOTE: looks like we currently do (opts: SignalAndInterval = {}) instead of using the ?, i.e. (opts?: SignalAndInterval). Our current way of setting a default feels nicer and has the same affact imo, but either should work.

There was some discussion of this [here](https://github.com/tablelandnetwork/js-tableland/pull/576#discussion_r1280864446). Personally I prefer the `?` for objects that don't have optional params. However, if we implement a true `opts` object on the top-level API (like [here](https://github.com/tablelandnetwork/js-tableland/commit/ef8f74f7acb51709d02f1c3dada2f69a21b55dd1)), then it _would_ have optional values and using the default pattern (` = {}`) makes sense to me, but internal methods that take `controller` would keep the `?` pattern because that object doesn't have optional params.

> I have a thought after looking at this with fresh eyes. Regardless if we go with per chain default timeouts, it would be nice to be able to set a global timeout via an https://github.com/tablelandnetwork/tableland-js/pull/36 type of interface.
This way, as a dev, I can decide how often my default polling should happen and how long to wait before stopping the polling.

Sounds good! Did you mean to link back to this PR? Should this come in this PR, or as a follow on?

dtbuchholz:
@sanderpick sgtm, will take a stab. and wrt the `overrideDefaults`, i _think_ it's referring to [this](https://github.com/tablelandnetwork/tableland-js/blob/19582eb4fea10388d2659dacdf5bd151cf73c2fe/packages/sdk/src/helpers/chains.ts#L120).

brunocalza:
> > According to the [current config ](https://github.com/tablelandnetwork/go-tableland/blob/main/docker/deployed/testnet/api/config.json#L84), we're waiting for at least one block to consider the tx final. and the ethereum block time is 12s, so I guess that's kind of expected. Were we experiencing faster times before?
> 
> @brunocalza ah okay that makes sense (i dont think they were faster before...just never really noticed it). out of curiosity, what's the a reason we impose the 1 block depth for sepolia and mumbai when arb/op goerli use `0`? do sepolia/mumbai commonly have reorgs or something?

i think it's because reorgs can't happen on L2s

joewagner:
> Sounds good! Did you mean to link back to this PR? Should this come in this PR, or as a follow on?

Oops, no I meant to link the `overideDefaults` method.  https://github.com/tablelandnetwork/tableland-js/blob/19582eb4fea10388d2659dacdf5bd151cf73c2fe/packages/sdk/src/helpers/chains.ts#L120

sanderpick:
Sweet, @dtbuchholz feel free to try and implement something like that here

packages/sdk/src/helpers/await.ts
```diff
@@ -1,56 +1,130 @@
...
+  pollingInterval: number = 1500
+): PollingController {
+  const controller = new AbortController();
+  const timeoutId = setTimeout(function () {
+    controller.abort();
+  }, timeout);
+  return {
+    signal: controller.signal,
+    abort: () => {
+      clearTimeout(timeoutId);
```
dtbuchholz, line 106: Note: the only change I made here from Sander's original code was the addition of this `clearTimeout(timeoutId)` line, to make sure that it's cleared if an abort occurs.

packages/sdk/src/helpers/chains.ts
```diff
@@ -11,18 +13,28 @@ export type ChainName = keyof TablelandNetworkConfig;
...
+ * @property pollingInterval The validator polling interval for the chain.
+ * @property [key: string] Any additional properties.
  */
 export interface ChainInfo {
   chainName: ChainName;
   chainId: number;
   contractAddress: string;
   baseUrl: string;
+  pollingTimeout: number;
+  pollingInterval: number;
```
dtbuchholz, line 30: I figured that since `overrideDefaults` is already storing validator-specific and chain information, we could just include the validator polling info there. If needed, I can separate it out into its own interface.

packages/sdk/src/helpers/chains.ts
```diff
@@ -11,18 +13,28 @@ export type ChainName = keyof TablelandNetworkConfig;
...
   [key: string]: ChainInfo[keyof ChainInfo];
 }
 
-// We simply pull this automatically from @tableland/evm to avoid keeping track seperately here.
+// We simply pull this automatically from @tableland/evm to avoid keeping track separately here.
 const entries = Object.entries(proxies) as Array<[ChainName, string]>;
 const mapped = entries.map(([chainName, contractAddress]) => {
+  // @ts-expect-error this imported object's values are always a string
```
dtbuchholz, line 37: Because the [`TablelandNetworkConfig`](https://github.com/tablelandnetwork/evm-tableland/blob/main/network.ts#L1) can now have types of `string | number`, the compiler was complaining here with the `new URL` constructor and `baseURIs[chainName]`. I tried a few different fixes like checking types & throwing etc, but nothing worked. I think it's safe to ignore since we know that `baseURIs` will always be strings, though.

packages/sdk/src/helpers/chains.ts
```diff
@@ -31,9 +43,20 @@ const mapped = entries.map(([chainName, contractAddress]) => {
       .filter((v) => v !== "")
       .pop() /* c8 ignore next */ ?? ""
   );
+  // Use per-chain validator polling timeout period
+  const pollingTimeout = validatorPollingTimeouts[chainName];
+  // Default to 1500ms polling interval, except for Filecoin due to long block times
+  const pollingInterval = chainName.includes("filecoin") ? 5000 : 1500;
```
dtbuchholz, line 49: For each chain, we take a chain-specific validator polling timeout as defined [here](https://github.com/tablelandnetwork/evm-tableland/blob/469f11ab10682eee8ccbfd86a7e8784f176df4dd/network.ts#L75C4-L75C4), and then we can use that polling timeout by default in other methods. For the interval, I've set the frequency to 1500ms for all chains—except for Filecoin.

Since it can take a few minutes for the validator to materialize FIL data, the standard interval is, instead, 5 seconds...it could be even longer, if needed.

packages/sdk/src/helpers/chains.ts
```diff
@@ -112,6 +140,18 @@ export function getBaseUrl(chainNameOrId: ChainName | number): string {
   return getChainInfo(chainNameOrId).baseUrl;
 }
 
+/**
+ * Create a polling controller with chain-specific timeout & interval.
+ * @param chainNameOrId The requested chain name.
+ * @returns A {@link PollingController} with standard timeout & interval per-chain.
+ */
+export function getChainPollingController(
```
dtbuchholz, line 148: This method makes it easy to use the chain-specific validator polling timeouts/intervals. It'll create a new polling controller with those chain-specific settings.

packages/sdk/src/statement.ts
```diff
@@ -32,6 +32,14 @@ import {
 
 export { type ValuesType, type Parameters, type ValueOf, type BaseType };
 
+/**
+ * Options for `all`, `first`, `run`, and `raw` methods.
+ * @property controller An optional {@link PollingController} used to control receipt polling behavior.
+ */
+export interface Options {
```
dtbuchholz, line 39: Here, we're now using a top-level options API, which is passed to the other statement methods below. It just stores the polling controller.

packages/sdk/src/statement.ts
```diff
@@ -131,29 +139,21 @@ export class Statement<S = unknown> {
...
-  ): Promise<Result<T>>;
-  async all<T = S, K extends keyof T = keyof T>(
-    colName: K,
-    opts?: SignalAndInterval
-  ): Promise<Result<T[K]>>;
-  async all<T = S, K extends keyof T = keyof T>(
-    colName?: K,
-    opts: SignalAndInterval = {}
-  ): Promise<Result<T | T[K]>> {
+  async all<T = S>(opts: Options = {}): Promise<Result<T>> {
```
dtbuchholz, line 156: Here's an example of how the options are passed.

Note that I had started down the path of D1 API conformance in this PR, but I ended up [creating a separate branch for this work](https://github.com/tablelandnetwork/tableland-js/pull/63). Looks like I removed some stuff on accident. Thus, the removed code _might_ need to be added back, unless that PR gets merged into this one.

packages/sdk/src/statement.ts
```diff
@@ -191,22 +185,22 @@ export class Statement<S = unknown> {
    * Instead it returns the object directly. If the query returns no
    * rows, then first() will return null.
    * @param colName If provided, filter results to the provided column.
-   * @param opts Additional options to control execution.
+   * @param opts An optional object used to control behavior, see {@link Options}
    */
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  async first<T = S, K extends keyof T = keyof T>(): Promise<T>;
+  async first<T = S, K extends keyof T = keyof T>(opts?: Options): Promise<T>;
```
dtbuchholz, line 191: Random TS learning—you can't use `opts: Options = {}` until the implementation, so the overloads must use `opts?: Options`

packages/sdk/src/validator/receipt.ts
```diff
@@ -83,6 +84,7 @@ export async function pollTransactionReceipt(
       throw err;
     }
   };
-  const receipt = await getAsyncPoller(fn, interval, signal);
+  const control = controller ?? getChainPollingController(params.chainId);
```
dtbuchholz, line 87: This is really the only place where `getChainPollingController` is used to wait longer on specific chains. It's what fixes the issue where things like the Studio or CLI abort early and never see the materialized table on networks like Sepolia or Filecoin.

packages/sdk/test/statement.test.ts
```diff
@@ -351,42 +354,38 @@ SELECT * FROM 3.14;
...
     test("when extracting a column from multiple rows", async function () {
-      const stmt = db.prepare(`SELECT * FROM ${tableName};`);
-      const { results, meta, error } = await stmt.all<any>("counter");
+      const stmt = db.prepare(`SELECT counter FROM ${tableName};`);
+      const { results, meta, error } = await stmt.all<any>();
       strictEqual(meta.txn, undefined);
       strictEqual(error, undefined);
       assert(meta.duration != null);
-      deepStrictEqual(results, [1, 2, 3, 4]);
+      deepStrictEqual(results, [
```
dtbuchholz, line 383: As noted, I started down the path of D1 API conformance and looks like I muddied up a bit here. So, the `colName` can't be passed to `all()` anymore. Again, I can revert this change here if we want to keep things simple and/or still want that feature. The [PR](https://github.com/tablelandnetwork/tableland-js/pull/63) with D1 changes, for reference.

packages/sdk/src/statement.ts
sanderpick: sounds good. I don't quite follow the point about removed code, but if easier, you could just target this branch. generally speaking though, its easier to keep PRs small and specific.

sanderpick (approved):
Very nice... thanks for adding all the extra tests.

packages/sdk/test/statement.test.ts
dtbuchholz: yeah, that makes sense to me.

also tagging @joewagner to help address one of your questions above about the removed code since he was helping with some cleanups for the D1 conformance. do the changes in [this comment](https://github.com/tablelandnetwork/tableland-js/pull/36/files/93d2adee5704935604d88536538cce4e4006f70e#r1375146455) make sense? namely, this is the current SDK code, before this PR and any D1 conformance:

```ts
async all<T = S, K extends keyof T = keyof T>(
  colName?: undefined,
  opts?: SignalAndInterval
): Promise<Result<T>>;
async all<T = S, K extends keyof T = keyof T>(
  colName: K,
  opts?: SignalAndInterval
): Promise<Result<T[K]>>;
async all<T = S, K extends keyof T = keyof T>(
  colName?: K,
  opts: SignalAndInterval = {}
): Promise<Result<T | T[K]>
```
and with the changes in this PR & D1 conformance (#63), we dropped `colName`, and it now simply looks like [this](https://github.com/tablelandnetwork/tableland-js/pull/63/files#r1375141792):
 
```ts
async all<T = Record<string, S>>(opts: Options = {}): Promise<Result<T>>
```

similarly, the `first` method currently looks like this, before any changes from those PRs:

```ts
async first<T = S, K extends keyof T = keyof T>(): Promise<T>;
async first<T = S, K extends keyof T = keyof T>(
  colName: undefined,
  opts?: SignalAndInterval
): Promise<T>;
async first<T = S, K extends keyof T = keyof T>(
  colName: K,
  opts?: SignalAndInterval
): Promise<T[K] | null>;
async first<T = S, K extends keyof T = keyof T>(
  colName?: K,
  opts: SignalAndInterval = {}
): Promise<T | T[K] | null>
```
and after these PR changes:

```ts
async first<T = Record<string, S>>(opts?: Options): Promise<T | null>;
async first<T = S, K extends keyof T = keyof T>(
  colName: K,
  opts?: Options
): Promise<T[K] | null>;
async first<T = S, K extends keyof T = keyof T>(
  colName?: K,
  opts: Options = {}
): Promise<T | T[K] | null> 
```
the main callout with `first` is that i dropped the override with `colName: undefined` (primarily because chatgpt didn't see why it was useful). and if no `colName` is specified at all, we use the Record/return types from the new D1 API.

joewagner: Making sure I'm on the right page here, are we trying to match these types? https://github.com/cloudflare/workerd/blob/main/types/defines/d1.d.ts

Without trying it, would the `colName: undefined` let you call first like this: `db.prepare(...).first(colName, options);`
Where colName is of type `string | undefined`?

dtbuchholz: correct, i was just trying to match those types. 

so, it looks like we _do_ need to keep the `colName: undefined` overload...i'll revert/add it back in here and just clean up this PR to make things easier / use #63 for more test cases.

what's interesting is that if i don't specify the `opts`/controller, we dont need the `colName: undefined` overload; it'll execute the same, whether or not that overload definition is included. but if we do use `opts`/controller, then it fails, and it only works if i add the `colName: undefined` overload back in there.

```ts
let c: string | undefined;
c = undefined;
let row = await stmt.first<any>(c); // works
row = await stmt.first<any>(c, { controller }); // fails
// error: Argument of type 'undefined' is not assignable to parameter of type 'string | number | symbol'. 
// The call would have succeeded against this implementation, but implementation signatures of overloads are not externally visible.
```

joewagner: @dtbuchholz  I'm ok adding it back, or not, up to you.
To put it all in one place... If we can support the following signatures, that would make sense to me:
```
// needed for ORM
await stmt.first();
// needed for ORM, i think?
await stmt.first(col);
// needed for users who want to modify there timeouts and specify a col
await stmt.first(col, opts); // NOTE: I don't know if we **must** allow `col` to be undefined, i'm ok either way.
// needed for users who want to modify there timeouts
await stmt.first(opts);
```

PR #39, opened 2023-09-27
Export Schema and Column types so they can be used externally

asutula:
Can't use the `Schema` type without this.

I was trying to use `type Schema = Table["schema"]` and then exporting things that use `Schema`, but of course got a ts error that the SDK's internal `Schema` type can't be named, which makes sense since it wasn't exported.

Figure we might as well export `Column` as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11969,24 +11969,24 @@
       "dev": true
     },
     "node_modules/typedoc": {
-      "version": "0.24.8",
-      "resolved": "https://registry.npmjs.org/typedoc/-/typedoc-0.24.8.tgz",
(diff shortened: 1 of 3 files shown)


asutula:
No idea why the tests would be failing. Any ideas what's up?

package-lock.json
asutula, line 1: This is a result of just running `npm install` in the root of the project.

joewagner:
> No idea why the tests would be failing. Any ideas what's up?

Yeah, this is a known bug https://github.com/tablelandnetwork/tableland-js/issues/40.  If you see that the tests are failing with `Error: deploying registry contract failed, cannot start network`, then you have to re-run failed jobs.

package-lock.json
joewagner: Looks like all the updates are needed, so this works well.

joewagner (approved):

carsonfarmer:
Nice!

asutula:
Rerunning tests again... 🤞🏼

carsonfarmer (approved):

PR #48, opened 2023-10-02
(sdk) add helper to create studio project aliases

joewagner:
## Overview
This PR exposes a helper function that takes a Studio project ID and returns an aliases Object that can be used to instantiate a DB which uses the given project's table names.

## Details
The end goal here is to make the SDK _feel_ like a normal database, specifically table names don't need their universally unique suffixes.  A dev can create a project in the Studio, then simply instantiate Databases with the projectId.
e.g. `new Database({ projectId: "123-xyz" });`

### This is a first draft and there's a few things to consider:

I ended up calling the config param `projectId`.  The names `environmentId`, and `databaseId` are being discussed as well.  For this draft I landed on `projectId` for the following reasons:
- The term Environment doesn't exist in the Studio and using it seemed like I was adding unecessary complexity.
- The term Database seems overloaded, and also doesn't exist in the Studio.
- The project page is where people will end up looking for the projectId.
- The existing api endpoint that returns the name mapping we want is `deployments.projectDeployments`, which takes a `projectId` argument.

I did write some tests to make development of the feature easier, but they only work when the tests can connect to the studio with pre-populated data.  I used a personal test wallet and created projects on the current production studio, but I'm `skip`ping the tests here because the studio instance url and data might change as we develop and having access to the wallet in github ci is not trivial.  For now the tests only serve as an example usage.

The `@tableland/studio-client`  package is going to need to be published.  As an alternative to this PR, we could export this feature as a function in the `@tableland/studio-client` package. thoughts?

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tableland/sdk",
-  "version": "4.5.3-dev.0",
+  "version": "4.6.0-pre.0",
   "description": "A database client and helpers for the Tableland network",
   "publishConfig": {
     "access": "public"
--- a/packages/sdk/src/helpers/config.ts
+++ b/packages/sdk/src/helpers/config.ts
@@ -1,3 +1,4 @@
+import { api as studioApi } from "@tableland/studio-client";
 import { type WaitableTransactionReceipt } from "../registry/utils.js";
 import { type FetchConfig } from "../validator/client/index.js";
 import { type ChainName, getBaseUrl } from "./chains.js";
@@ -112,6 +113,40 @@ export function jsonFileAliases(filepath: string): AliasesNameMap {
     write: async function (nameMap: NameMapping) {
       const fs = await getFsModule();
       fs.writeFileSync(filepath, JSON.stringify(nameMap));
+    },
+  };
+}
+
+// NOTE: In the future we may need to use `environmentId` instead of `projectId`, but
+//  there is currently no concept of an environment for a user, and the api doesn't
+//  support querying for deployments based on environment.
+export function studioAliases(
+  projectId: string,
+  apiUrl?: string
+): AliasesNameMap {
+  const api = studioApi({
+    url: apiUrl,
+  });
+  const loadMap = async function (): Promise<void> {
+    const res = await api.deployments.projectDeployments.query({ projectId });
+
+    _map = {};
+    // map the response to a `NameMapping` Object
+    // { tokenId: string; tableId: string; tableName: string; environmentId: string; chainId: number; blockNumber: number | null; txnHash: string | null; createdAt: string; }
+    res.forEach(function (row) {
+      _map[row.tableName.split("_").slice(0, -2).join("_")] = row.tableName;
+    });
+  };
+
+  let _map: NameMapping;
+  return {
+    read: async function (): Promise<NameMapping> {
+      if (typeof _map === "undefined") await loadMap();
+
+      return _map;
+    },
+    write: async function () {
+      throw new Error("cannot create project tables via studio sdk aliases");
     },
   };
 }
--- a/packages/sdk/test/aliases.test.ts
+++ b/packages/sdk/test/aliases.test.ts
@@ -2,13 +2,15 @@ import url from "node:url";
 import path from "node:path";
 import fs from "node:fs";
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
-import { strictEqual, rejects } from "assert";
+import { deepStrictEqual, equal, strictEqual, rejects } from "assert";
 import { describe, test } from "mocha";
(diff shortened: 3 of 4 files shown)


packages/sdk/src/helpers/config.ts
```diff
@@ -116,6 +117,40 @@ export function jsonFileAliases(filepath: string): AliasesNameMap {
   };
 }
 
+// NOTE: In the future we may need to use `environmentId` instead of `projectId`, but
+//  there is currently no concept of an environment for a user, and the api doesn't
+//  support querying for deployments based on environment.
+export function studioAliases(
```
joewagner, line 123: This is the meat of the feature.  The studio project name map is loaded and used for read and insert only.

packages/sdk/src/helpers/config.ts
```diff
@@ -116,6 +117,40 @@ export function jsonFileAliases(filepath: string): AliasesNameMap {
...
+
+  let _map: NameMapping;
+  return {
+    read: async function (): Promise<NameMapping> {
+      if (typeof _map === "undefined") await loadMap();
+
+      return _map;
+    },
+    write: async function () {
+      throw new Error("cannot create project tables via studio sdk aliases");
```
joewagner, line 149: The write function may never be used since creating tables must be done via Studio

packages/sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tableland/sdk",
-  "version": "4.5.3-dev.0",
+  "version": "4.6.0-pre.0",
```
joewagner, line 3: We will need a minor version bump

packages/sdk/test/aliases.test.ts
```diff
@@ -232,4 +234,76 @@ describe("aliases", function () {
...
   });
+
+  describe.skip("studio based aliases", function () {
+    // TODO: these values are set per the actual production studio. It's tempting
+    //       to sandbox this, but doing so would add significant complexity.
+    // TODO: move the argument values to env vars.
+    const aliases = studioAliases(
+      "6f254b66-d9cf-482b-a4b9-76cfe5eb2f19",
+      "https://studio-neon.vercel.app/"
+    );
```
joewagner, line 245: usage with the SDK is fairly straight forward.  Just provide a projectId and the URL of the studio.  We can remove the requirement for studio url once we have a permanent url.

packages/sdk/src/helpers/config.ts
```diff
@@ -1,3 +1,4 @@
+import { api as studioApi } from "@tableland/studio-client";
```
joewagner, line 1: The `@tableland/studio-client` package is not published yet.  Anyone wanting to try out this feature will have to `npm link`.  Linking between to two monorepos is tricky.  I can help with that via voice if needed.

packages/sdk/src/helpers/config.ts
```diff
@@ -116,6 +117,40 @@ export function jsonFileAliases(filepath: string): AliasesNameMap {
...
+//  support querying for deployments based on environment.
+export function studioAliases(
+  projectId: string,
+  apiUrl?: string
+): AliasesNameMap {
+  const api = studioApi({
+    url: apiUrl,
+  });
+  const loadMap = async function (): Promise<void> {
+    const res = await api.deployments.projectDeployments.query({ projectId });
```
asutula, line 131: This only works now because there is a single environment called `default`. Once there are multiple environments, you'll end up with duplicate aliases in the map. Actually they will just overwrite each other and the map will end up with a mix of tables from different environments.

The root of this problem is using `projectId` as the parameter. We should have the user pass in the environment id. We can call it something opaque sounding like `key` or whatever since "environment id" isn't a user facing concept yet. 

If we do that, we can create a new api function to return deployments for an environment id, then the map will be accurate.

asutula (commented):
This is looking like a good start. I left some critical comments. 

My opinion is that we should not actually do this work in the SDK for now. It's adding too much complexity with having to `npm link` monorepos for development, publish the `@tableland/studio-client` package, etc. Plus, the current API for this helper is very alpha right now since we haven't properly introduced environments (or databases if we call them that). For those reasons, I vote to just create a tiny package in the Studio monorepo that publishes this helper.

joewagner:
> My opinion is that we should not actually do this work in the SDK for now. It's adding too much complexity with having to `npm link` monorepos for development, publish the `@tableland/studio-client` package, etc. Plus, the current API for this helper is very alpha right now since we haven't properly introduced environments (or databases if we call them that). For those reasons, I vote to just create a tiny package in the Studio monorepo that publishes this helper.

I like the idea of moving the helper to the studio repo.  What about putting it in the `client` package?

asutula:
Sounds good to me.

joewagner:
replaced by: https://github.com/tablelandnetwork/studio/pull/85

PR #51, opened 2023-10-12
Expose a Local Tableland method that restarts the validator

joewagner:
## Overview
Hardhat has an RPC API that contains methods to take a snapshot of the chain state, and then revert to a given state.
Docs are here: https://hardhat.org/hardhat-network/docs/reference#evm_snapshot
This is useful for testing contracts, but the local tableland network becomes out of sync when this happens because the tables materialized by the Validator are no longer valid.
This PR enables local tableland users who want to snapshot and revert to do so in a way that keeps the Validator up to date with the reverted chain state.

## Details
The mechanism used here to re-sync the validator with the chain is to simple shutdown and restart the validator.  The methods needed to accomplish that haven't been public, so this PR simply does a small refactor and exposes a new method called `restartValidator`.

### DIFF
--- a/packages/local/src/main.ts
+++ b/packages/local/src/main.ts
@@ -165,6 +165,49 @@ class LocalTableland {
       );
     }
 
+    await this.#_startValidator();
+    await this.#_setReady();
+
+    if (this.silent as boolean) return;
+
+    console.log("\n\n******  Tableland is running!  ******");
+    console.log("             _________");
+    console.log("         ___/         \\");
+    console.log("        /              \\");
+    console.log("       /                \\");
+    console.log("______/                  \\______\n\n");
+    console.log("Using Configuration:\n" + JSON.stringify(config, null, 4));
+    console.log("\n\n*************************************\n");
+  }
+
+  // note: Tests are using sinon to stub this method. Because typescript compiles ecmascript
+  //       private features, i.e. hash syntax, in a way that does not work with sinon we must
+  //       use the ts private modifier here in order to test the failure to deploy the registry.
+  private _deployRegistry(): void {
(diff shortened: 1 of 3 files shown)


packages/local/src/main.ts
```diff
@@ -165,6 +165,49 @@ class LocalTableland {
       );
     }
 
+    await this.#_startValidator();
```
joewagner, line 168: I moved the validator initialization logic into a function that can be called multiple times.

packages/local/src/main.ts
```diff
@@ -165,6 +165,49 @@ class LocalTableland {
...
+    await this.#_setReady();
+
+    if (this.silent as boolean) return;
+
+    console.log("\n\n******  Tableland is running!  ******");
+    console.log("             _________");
```
joewagner, line 174: The diff here looks a bit weird because I moved the validator init code without changing it. Because of that this is added and removed.

packages/local/src/main.ts
```diff
@@ -273,6 +277,11 @@ class LocalTableland {
     return await prom;
   }
 
+  async restartValidator(): Promise<void> {
+    await this.shutdownValidator();
+    await this.#_startValidator();
```
joewagner, line 282: We already had a `shutdownValidator` method, just needed to add the `_startValidator`

joewagner (commented):
@carsonfarmer This is the change I came up with for the fix Dimo is asking for in Discord.
If this looks good/makes sense to you, I'll publish a pre version today so they can try it out.

avichalp (approved):
Looks good!

sanderpick (approved):
Very nice!

PR #59, opened 2023-10-24
fix: filter out `optimism-goerli-staging` from chains

### LINKED ISSUE #58: Fix: remove optimism-goerli-staging from SDK chains
## Summary

If you deploy a table using Optimsim Goerli, it will default to the `optimism-goerli-staging` contract instead of `optimism-goerli`. This is a bug since the staging contract should not be used by any developers externally. See discussion [here](https://linear.app/tableland/issue/STU-176/web-optimism-goerli-staging-is-used-instead-of-optimism-goerli) for more details.

## Details

See this [line](https://github.com/tablelandnetwork/tableland-js/blob/19582eb4fea10388d2659dacdf5bd151cf73c2fe/packages/sdk/src/helpers/chains.ts#L44)—here (or in the preceding code) we must filter out the "staging" keyword. The CLI does this with the following, so something similar must be put into place:

```
export const getChains = function (): typeof helpers.supportedChains {
  return Object.fromEntries(
    Object.entries(helpers.supportedChains).filter(
      ([name]) => !name.includes("staging")
    )
  ) as Record<helpers.ChainName, helpers.ChainInfo>;
};
```

<sub>From [SyncLinear.com](https://synclinear.com) | [JST-35](https://linear.app/tableland/issue/JST-35/fix-remove-optimism-goerli-staging-from-chains)</sub>

dtbuchholz:
## Summary

If you try to create a table on OP Goerli, it actually uses the `optimism-goerli-staging` network instead of `optimism-goerli`. This filters out the "staging" network and adds tests accordingly. Also, I squeezed in a dep bump for `@tableland/sdk`. Closes #58.

## Details

There's a simple one liner added to `helpers/chains.ts`:
```ts
const filtered = mapped.filter(
  ([chainName]) => chainName !== "optimism-goerli-staging"
);
```

This is then used in the `supportedChains` object, which gets exported and is used in clients like the CLI. Note that the CLI had to implement similar logic to what's described above. The change in the SDK will allow us to remove that check in the CLI, but it's not required at the moment; everything should still work as expected in the CLI.

Wrt to the `@tableland/sdk@^4.5.3-dev.0` dep bump, I also deleted `package-lock.json` and reinstalled the overall deps in root. It seems that lerna was having issues with symlinks, and deleting the lock file & updating the packages seemed to fix it. I think lerna might require that each package is using the latest version—e.g., in `packages/cli`, if I don't use `4.5.3-dev.0` but use any earlier SDK version, I cannot set up a symlink successfully.

## How it was tested

Added a test in `chains.test.ts` to make sure that `getChainId("optimism-goerli-staging")` will throw an error for `cannot use unsupported chain`.

I also set up a symlink to the Studio, and I successfully created a table on the _correct_ OP Goerli network.

## Checklist:

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -54,7 +54,7 @@
   },
   "dependencies": {
     "@ensdomains/ensjs": "3.0.0-alpha.64",
-    "@tableland/sdk": "^4.4.2",
+    "@tableland/sdk": "^4.5.3",
     "@tableland/sqlparser": "^1.3.0",
     "cli-select-2": "^2.0.0",
     "cosmiconfig": "^8.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -12656,7 +12656,7 @@
       "license": "MIT AND Apache-2.0",
       "dependencies": {
         "@ensdomains/ensjs": "3.0.0-alpha.64",
-        "@tableland/sdk": "^4.4.2",
+        "@tableland/sdk": "^4.5.3",
         "@tableland/sqlparser": "^1.3.0",
         "cli-select-2": "^2.0.0",
         "cosmiconfig": "^8.0.0",
@@ -12681,17 +12681,6 @@
         "mock-stdin": "^1.0.0",
         "sinon": "^15.0.1",
         "tempy": "^3.0.0"
-      }
-    },
-    "packages/cli/node_modules/@tableland/sdk": {
-      "version": "4.5.2",
-      "resolved": "https://registry.npmjs.org/@tableland/sdk/-/sdk-4.5.2.tgz",
-      "integrity": "sha512-fshBi0tyRGhLQY3C23XlGpbdrD2bPGiCFnOgGf7W83S6vB1aELM3yTMaVuQslmkwOcTW4bzSkzlq4TMPd0LR4A==",
-      "dependencies": {
-        "@async-generators/from-emitter": "^0.3.0",
-        "@tableland/evm": "^4.3.0",
-        "@tableland/sqlparser": "^1.3.0",
-        "ethers": "^5.7.2"
       }
     },
     "packages/cli/node_modules/ansi-escapes": {
@@ -12910,7 +12899,7 @@
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
-        "@tableland/sdk": "^4.4.2",
+        "@tableland/sdk": "^4.5.3",
         "@tableland/validator": "^1.8.1",
         "cross-spawn": "^7.0.3",
         "enquirer": "^2.3.6",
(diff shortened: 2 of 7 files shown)


dtbuchholz:
@joewagner i made a small adjustment to the SDK to fix a bug that was uncovered during ETHOnline while someone was using the Studio. it removes `optimism-goerli-staging` from the available chains, and i included some dep chores to resolve lerna symlink issues. lmk if the deps aspect isn't ideal, and i can remove em. cc @asutula

packages/cli/src/commands/write.ts
```diff
@@ -111,9 +111,8 @@ export const handler = async (argv: Arguments<Options>): Promise<void> => {
             norm.tables[0]
           );
 
-          const { tableId } = await globalThis.sqlparser.validateTableName(
-            name
-          );
+          const { tableId } =
+            await globalThis.sqlparser.validateTableName(name);
```
joewagner, line 115: Looks like the prettier settings are changed.  Is this something to do with the deleted package lock?

packages/cli/package.json
```diff
@@ -54,7 +54,7 @@
   },
   "dependencies": {
     "@ensdomains/ensjs": "3.0.0-alpha.64",
-    "@tableland/sdk": "^4.4.2",
+    "@tableland/sdk": "^4.5.3-dev.0",
```
joewagner, line 57: Let's bump the sdk to the next appropriate version.

joewagner:
@dtbuchholz this looks good. We might as well put the sdk at an official release number.  Looks like something got messed up during the last publish and npm has the sdk's `latest` version set at `4.5.3-dev.0`, which should be the `next` version.
I can guess how that happened (probably my mistake) but let's fix it with this pr.  I think version `4.5.3` makes sense here since it's considered a bug.

asutula (approved):

packages/cli/src/commands/write.ts
dtbuchholz: yeah, looks like it. i dropped that commit, and the formatting went back to normal. i'll just ignore any of the lerna-related things i included here...it _might_ not be the root cause, anyways, after some additional testing...

dtbuchholz:
> I think version 4.5.3 makes sense here since it's considered a bug.

@joewagner i bumped the versions in the cli & local packages to 4.5.3. but since it doesn't exist, the tests etc obviously fail here. is that fine, or should i move the dep bump to a separate PR?

joewagner:
> @joewagner i bumped the SDK versions in the cli & local packages to 4.5.3. but since it doesn't exist, the tests etc obviously fail here. is that fine, or should i move the dep bump to a separate PR?

> edit: i'll also have to bump the `package/sdk`'s `package.json` version to `4.5.3` too, right? or does the publish GH action somehow handle that? (this change is not included yet, but i'll wait for your input above before i do anything there.)

You should be able to set the version to `4.5.3`, then update the dependencies in the other packages and as long as the build and test works this PR is good to go.  Once merged someone(I can do it, or we can go through it together in voice), can run the lerna publish flow described here: https://www.notion.so/textile/Lerna-Publish-Playbook-55fab12115fd4506926ee1a0556ad63a

joewagner:
@dtbuchholz I accidentally edited your comment instead of replying...
I tried to put it back how it was, but I'm not totally sure.
I really don't understand why github lets you edit someone else's comment

dtbuchholz:
> @dtbuchholz I accidentally edited your comment instead of replying... I tried to put it back how it was, but I'm not totally sure. I really don't understand why github lets you edit someone else's comment

@joewagner no worries...that's such a dumb feature lol

> Once merged someone(I can do it, or we can go through it together in voice), can run the lerna publish flow described here: https://www.notion.so/textile/Lerna-Publish-Playbook-55fab12115fd4506926ee1a0556ad63a

i'm down to take a stab! i'll ping you on discord since i know timezones are a bit tough rn

joewagner (approved):

PR #60, opened 2023-10-24
(sdk) Add alias to file instead of overwriting

joewagner:
This current aliases config helper will overwrite any existing aliases file when you create a new table.  This was an oversight during initial development.  This PR changes the behaviour to add to the file as new tables are created.


dtbuchholz (approved):
🚀

PR #63, opened 2023-10-28
refactor: D1 API conformance

dtbuchholz:
## Summary

The D1 API changed, so our API is no longer compatible. This introduces some breaking changes since some of the statement & database methods' parameters, generic types, and returned types have to be altered for conformance purposes.

## Details

For reference, [these are the old](https://github.com/cloudflare/workerd/commit/72d6915e19fb56de8373144bf922afa253b378cc) APIs, and [these are the new ones](https://github.com/cloudflare/workerd/blob/main/types/defines/d1.d.ts). The interfaces the statement methods must conform to are the following—notice the generic type `T = unknown` is now `T = Record<string, unknown>` in most of these:

```ts
// old
bind(...values: any[]): D1PreparedStatement;
first<T = unknown>(colName?: string): Promise<T>;
run<T = unknown>(): Promise<D1Result<T>>;
all<T = unknown>(): Promise<D1Result<T>>;
raw<T = unknown>(): Promise<T[]>;

// new
bind(...values: unknown[]): D1PreparedStatement;
first<T = unknown>(colName: string): Promise<T | null>;
first<T = Record<string, unknown>>(): Promise<T | null>;
run<T = Record<string, unknown>>(): Promise<D1Result<T>>;
all<T = Record<string, unknown>>(): Promise<D1Result<T>>;
raw<T = unknown[]>(): Promise<T[]>;
```

None of the database methods have changed, except for `exec`, which now has a custom return type:

```ts
// old
interface D1Result<T = unknown> {
  results?: T[];
  success: boolean;
  error?: string;
  meta: any;
}
exec<T = unknown>(query: string): Promise<D1Result<T>>;

// new
interface D1ExecResult {
  count: number;
  duration: number;
}
exec(query: string): Promise<D1ExecResult>;
```

So, a new `ExecResult` interface has been added for the `exec` method's return type, which includes the required `count` and `duration` as well as optional `txn` and `results` (see below).

From a DX perspective, the following changes should be noted. It wasn't clear if all of these changes were necessary as I was simply comparing and trying to go 1:1 to the new D1 API:
- Removed `colName` parameter for the statement methods, except `first`.
- Changed generic types for relevant statement methods to use `T = Record<string, S>` instead of `T = S`, where `S = unknown`.
- For `exec`, it no longer can use the same `Result` we had—a new `ExecResult` interface has been added, and a `wrapExecResult` properly formats the data for the `exec` return type.
  - This introduces one notable breaking change—since this interface only requires `count` and `duration`, the typical `meta` object cannot be required.
  - Thus, instead of a pattern for mutating queries like `await meta.txn?.wait()`, it looks like `await txn?.wait()`.

## How it was tested

Altered existing tests to take these changes into account. It primarily affected the `exec` method since it uses a new interface that doesn't contain `meta`.

Note: it looks like I also included a new test for `getChainPollingController`, which is more relevant to https://github.com/tablelandnetwork/tableland-js/pull/36.

## Checklist:

### DIFF
--- a/packages/sdk/src/database.ts
+++ b/packages/sdk/src/database.ts
@@ -1,6 +1,10 @@
 import { type NormalizedStatement } from "@tableland/sqlparser";
-import { type Result, type Runnable } from "./registry/index.js";
-import { wrapResult } from "./registry/utils.js";
+import {
+  type Result,
+  type ExecResult,
+  type Runnable,
+} from "./registry/index.js";
+import { wrapResult, wrapExecResult } from "./registry/utils.js";
 import {
   type Config,
   type AutoWaitConfig,
@@ -194,16 +198,14 @@ export class Database<D = unknown> {
   async exec<T = D>(
     statementStrings: string,
     controller?: PollingController
-  ): Promise<Result<T>> {
+  ): Promise<ExecResult<T>> {
     // TODO: Note that this method appears to be the wrong return type in practice.
     try {
       const { statements } = await normalize(statementStrings);
       const count = statements.length;
       const statement = this.prepare(statementStrings);
-      const result = await statement.run({ controller });
-      // Adds a count property which isn't typed
-      result.meta.count = count;
-      return result;
+      const result = await statement.run<T>({ controller });
+      return wrapExecResult(result, count);
     } catch (cause: any) {
       if (cause.message.startsWith("RUN_ERROR") === true) {
         throw errorWithCause("EXEC_ERROR", cause.cause);
--- a/packages/sdk/src/registry/utils.ts
+++ b/packages/sdk/src/registry/utils.ts
@@ -101,6 +101,7 @@ export function wrapResult<T = unknown>(
     meta,
     success: true,
     results: [],
+    error: undefined,
   };
   if (isTransactionReceipt(resultsOrReceipt)) {
     return { ...result, meta: { ...meta, txn: resultsOrReceipt } };
@@ -159,11 +160,13 @@ export interface Result<T = unknown> {
   /**
    * Whether the query or transaction was successful.
    */
-  success: boolean; // almost always true
+  success: true; // TODO: this is a bug in D1, but if we want to be compatible
+  //       we have to type it like this :<
+  //       https://github.com/cloudflare/workerd/issues/940
   /**
    * If there was an error, this will contain the error string.
    */
-  error?: string; // TODO: this changed to `never` in D1 API
+  error: undefined;
   /**
    * Additional meta information.
    */
--- a/packages/sdk/src/registry/utils.ts
+++ b/packages/sdk/src/registry/utils.ts
@@ -86,6 +86,12 @@ function isTransactionReceipt(arg: any): arg is WaitableTransactionReceipt {
   );
 }
 
+/**
+ * Wrap results for a Statement `run` or `all` call, or a Database `batch` call.
+ * @param resultsOrReceipt Either query results or the transaction receipt.
+ * @param duration Total client-side duration of the async call.
+ * @returns Wrapped results with metadata.
+ */
 export function wrapResult<T = unknown>(
   resultsOrReceipt: T[] | WaitableTransactionReceipt,
   duration: number
@@ -103,6 +109,28 @@ export function wrapResult<T = unknown>(
 }
 
 /**
+ * Wrap results for a Database `exec` call.
+ * @param result The result of the a {@link wrapResult} call, made by `exec` under the hood.
+ * @param count The count of executed statements.
+ * @returns Wrapped {@link ExecResult} with metadata and transaction
+ * receipt or query results.
+ */
+export function wrapExecResult<T = unknown>(
+  result: Result<T>,
+  count: number
+): ExecResult<T> {
+  const { duration } = result.meta;
+  const execResult: ExecResult<T> = {
+    count,
+    duration,
+  };
+  if (result.meta.txn != null) {
+    return { ...execResult, txn: result.meta.txn };
+  }
+  return { ...execResult, results: result.results };
+}
+
+/**
  * Metadata represents meta information about an executed statement/transaction.
  */
 export interface Metadata {
@@ -111,11 +139,11 @@ export interface Metadata {
    */
   duration: number;
   /**
-   * The optional transactionn information receipt.
+   * The optional transaction information receipt.
    */
   txn?: WaitableTransactionReceipt;
   /**
-   * Metadata may contrain additional arbitrary key/values pairs.
+   * Metadata may constrain additional arbitrary key/values pairs.
    */
   [key: string]: any;
 }
@@ -135,11 +163,26 @@ export interface Result<T = unknown> {
   /**
    * If there was an error, this will contain the error string.
    */
-  error?: string;
+  error?: string; // TODO: this changed to `never` in D1 API
   /**
    * Additional meta information.
    */
   meta: Metadata;
+}
+
+/**
+ * ExecResult represents the return result for executed Database statements via `exec()`.
+ */
+export interface ExecResult<T = unknown>
+  extends Pick<Metadata, "duration" | "txn"> {
+  /**
+   * The count of executed statements.
+   */
+  count: number;
+  /**
+   * The optional list of query results.
+   */
+  results?: T[];
 }
 
 export async function extractReadonly(
--- a/packages/sdk/src/statement.ts
+++ b/packages/sdk/src/statement.ts
@@ -232,7 +232,7 @@ export class Statement<S = unknown> {
    * Runs the query/queries, but returns no results. Instead, run()
    * returns the metrics only. Useful for write operations like
    * UPDATE, DELETE or INSERT.
-   * @param controller An optional object used to control behavior, see {@link Options}
+   * @param opts An optional object used to control behavior, see {@link Options}
    * @returns A results object with metadata only (results are null or an empty array).
    */
   async run<T = Record<string, S>>(opts: Options = {}): Promise<Result<T>> {
@@ -263,7 +263,7 @@ export class Statement<S = unknown> {
 
   /**
    * Same as stmt.all(), but returns an array of rows instead of objects.
-   * @param controller An optional object used to control behavior, see {@link Options}
+   * @param opts An optional object used to control behavior, see {@link Options}
    * @returns An array of raw query results.
    */
   async raw<T = S[]>(opts: Options = {}): Promise<Array<ValueOf<T>>> {
--- a/packages/sdk/src/statement.ts
+++ b/packages/sdk/src/statement.ts
@@ -12,6 +12,7 @@ import {
   checkWait,
   normalize,
 } from "./helpers/index.js";
+import { isPollingController } from "./helpers/await.js";
 import {
   type ExtractedStatement,
   type Result,
@@ -179,6 +180,11 @@ export class Statement<S = unknown> {
     }
   }
 
+  // Check if the first param seen by `first()` is an Options object
+  #checkIsValidOpts(opts: any): opts is Options {
+    return opts != null && isPollingController(opts.controller);
+  }
+
   /**
    * Executes a query and returns the first row of the results.
    * This does not return metadata like the other methods.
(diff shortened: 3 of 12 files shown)


packages/sdk/src/database.ts
```diff
@@ -194,16 +198,14 @@ export class Database<D = unknown> {
...
     try {
       const { statements } = await normalize(statementStrings);
       const count = statements.length;
       const statement = this.prepare(statementStrings);
-      const result = await statement.run({ controller });
-      // Adds a count property which isn't typed
-      result.meta.count = count;
-      return result;
+      const result = await statement.run<T>({ controller });
+      return wrapExecResult(result, count);
```
dtbuchholz, line 208: This return type had to be altered to use `ExecResult`, which looks like this:
```ts
interface ExecResult<T = unknown>
  count: number;
  duration: number;
  txn?: WaitableTransactionReceipt;
  results?: T[];
}
```
The D1 API only requires `count` and `duration`. This is where the DX component comes into play because everything _except_ this method can use the `meta.txn?.wait()` pattern, but this needs to do `txn?.wait()`. 

Instead of `ExecResult` having a `txn?` field, I debated making it a `meta?` field...but that means usage would look like `meta?.txn?.wait()`, and the `meta?` being a bit confusing imo.

dtbuchholz: Also, this method could be refactored, if needed. It just takes the result from `await statement.run(..)`, which means we're really calling `wrapResult` and then passing that value to `wrapExecResult`. But, I figured I'd keep it simple in case any of this isn't the right approach.

packages/sdk/src/registry/utils.ts
```diff
@@ -135,13 +163,28 @@ export interface Result<T = unknown> {
...
   /**
    * Additional meta information.
    */
   meta: Metadata;
 }
 
+/**
+ * ExecResult represents the return result for executed Database statements via `exec()`.
+ */
+export interface ExecResult<T = unknown>
```
dtbuchholz, line 176: Here's the `ExecResult`, for reference.

packages/sdk/src/statement.ts
```diff
@@ -153,7 +153,7 @@ export class Statement<S = unknown> {
    * Executes a query and returns all rows and metadata.
    * @param opts An optional object used to control behavior, see {@link Options}
    */
-  async all<T = S>(opts: Options = {}): Promise<Result<T>> {
+  async all<T = Record<string, S>>(opts: Options = {}): Promise<Result<T>> {
```
dtbuchholz, line 156: Generic type change for API conformance.

packages/sdk/src/statement.ts
```diff
@@ -187,13 +187,7 @@ export class Statement<S = unknown> {
...
    * @param opts An optional object used to control behavior, see {@link Options}
    */
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  async first<T = S, K extends keyof T = keyof T>(opts?: Options): Promise<T>;
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  async first<T = S, K extends keyof T = keyof T>(
-    colName: undefined,
-    opts?: Options
-  ): Promise<T>;
+  async first<T = Record<string, S>>(opts?: Options): Promise<T | null>;
```
dtbuchholz, line 190: Generic type change for API conformance.

Also, I wasn't sure if the lines I deleted were necessary with the addition of this method above, but I could be wrong so lmk if I should add any of those back in.

packages/sdk/src/statement.ts
```diff
@@ -241,7 +235,7 @@ export class Statement<S = unknown> {
    * @param controller An optional object used to control behavior, see {@link Options}
    * @returns A results object with metadata only (results are null or an empty array).
    */
-  async run(opts: Options = {}): Promise<Result<never>> {
+  async run<T = Record<string, S>>(opts: Options = {}): Promise<Result<T>> {
```
dtbuchholz, line 238: Generic type change for API conformance. Also, I'm not sure if removing that `never` is correct...I was just basing it off of the D1 API, which expects a `Result<T>` [here](https://github.com/cloudflare/workerd/blob/875a5171d1ec52d3bff63fa0e599975a769956f1/types/defines/d1.d.ts#L24).

packages/sdk/src/statement.ts
```diff
@@ -272,7 +266,7 @@ export class Statement<S = unknown> {
    * @param controller An optional object used to control behavior, see {@link Options}
    * @returns An array of raw query results.
    */
-  async raw<T = S>(opts: Options = {}): Promise<Array<ValueOf<T>>> {
+  async raw<T = S[]>(opts: Options = {}): Promise<Array<ValueOf<T>>> {
```
dtbuchholz, line 269: Generic type change, seen [here](https://github.com/cloudflare/workerd/blob/875a5171d1ec52d3bff63fa0e599975a769956f1/types/defines/d1.d.ts#L26).

packages/sdk/test/chains.test.ts
```diff
@@ -204,6 +204,66 @@ describe("chains", function () {
     });
   });
 
+  describe("getChainPollingController()", function () {
```
dtbuchholz, line 207: These are the tests that should probably be in https://github.com/tablelandnetwork/tableland-js/pull/36, so they're not related to the D1 conformance.

packages/sdk/test/database.test.ts
```diff
@@ -599,12 +599,12 @@ describe("database", function () {
     test("when executing mutations works and adds rows", async function () {
       const sql = `INSERT INTO ${tableName} (name, age) VALUES ('Bobby', 5);
       INSERT INTO ${tableName} (name, age) VALUES ('Tables', 6);`;
-      const { meta } = await db.exec(sql);
-      assert(meta.duration != null);
-      strictEqual(meta.count, 2);
-      assert(meta.txn != null);
+      const { duration, count, txn } = await db.exec(sql);
```
dtbuchholz, line 602: This shows how `exec` usage has changed. It feels _almost_ the same...just with top-level fields vs. nested in `meta`.

packages/sdk/test/thirdparty.test.ts
```diff
@@ -29,7 +29,14 @@ describe("thirdparty", function () {
   const signer = new NonceManager(baseSigner);
   const db = new Database({ signer });
 
-  describe("d1-orm", function () {
+  // NOTE: the d1-orm hasn't updated to the latest version of @cloudflare/workers-types
```
dtbuchholz, line 32: Lastly, the `d1-orm` test has been commented out since the ORM actually isn't bumped to the latest version of `@cloudflare/workers-types`

dtbuchholz:
(Also, looks like a test failed for `@tableland/local` due to the `exec` change, but I'll wait to clean that up to make sure these changes are the correct approach.)

joewagner (changes requested):
@dtbuchholz This looks pretty good so far.  I have a PR with some suggested additions to make the SDK D1 and typescript all work together here: https://github.com/tablelandnetwork/tableland-js/pull/65

joewagner:
In regard to versioning.
I think these changes would indicate a major bump right?  I'm not sure why d1-orm didn't do a major, but I guess since they are still version `0.x.x` they can break things without a major bump

joewagner:
> does lerna automatically use whatever code is present in other packages? e.g., for `local`, I didn't change up its deps at all, but a `local` test that used the `sdk` package's `exec` no longer worked due to the SDK's breaking change. in other words, the published version for `@tableland/sdk` is the listed dependency for `local`, but the `local` tests were using the SDK changes from this PR.

I think that's how lerna is designed to work.  If you need to have incompatible packages in the same commit I think you can pin an older version of `sdk` in the `local` package.lock, but I'm not sure.

packages/sdk/src/database.ts
dtbuchholz: cool, issue created [here](https://linear.app/tableland/issue/STU-164/web-aborted-signal-during-table-creation-in-ui)...should be a quick fix so i can take it on

dtbuchholz:
> I think these changes would indicate a major bump right? I'm not sure why d1-orm didn't do a major, but I guess since they are still version `0.x.x` they can break things without a major bump

@joewagner correct, we'll need a major version. and also correct—the d1-orm maintainer said that pre-major version, they're just keeping all breaking changes as minor versions.

packages/sdk/test/thirdparty.test.ts
```diff
@@ -263,8 +263,7 @@ describe("thirdparty", function () {
     type User = InferSelectModel<typeof user>;
     type NewUser = InferInsertModel<typeof user>;
 
-    // @ts-expect-error Tableland database & D1Database type assignment warning
-    const database = drizzle<User>(db, { schema: user });
+    const database = drizzle(db, { schema: { user } });
```
dtbuchholz, line 266: @joewagner i merged your PR to resolve the other `@ts-expect-error`s in this test (thanks for that!). and this last one turned out to be more of an error on my side, so it's been fixed, too.

packages/sdk/src/statement.ts
```diff
@@ -266,7 +266,7 @@ export class Statement<S = unknown> {
    * @param opts An optional object used to control behavior, see {@link Options}
    * @returns An array of raw query results.
    */
-  async raw<T = S[]>(opts: Options = {}): Promise<Array<ValueOf<T>>> {
+  async raw<T = unknown>(opts: Options = {}): Promise<Array<ValueOf<T>>> {
```
dtbuchholz, line 269: @joewagner oh actually, i have a question with this one. the [new API](https://github.com/cloudflare/workerd/blob/7314c771b6ca85bdc5d0b00592e889769d40eec7/types/defines/d1.d.ts#L26) uses:
```ts
raw<T = unknown[]>(): Promise<T[]>;
```
what's the reason you used `T = unknown` over `T = S[]` (where `S = unknown`)? is it more flexible to how we implement the underlying logic or something like that?

packages/sdk/src/statement.ts
dtbuchholz: As discussed in #36, I've added back the overload that has `colName: undefined`. (I think I was a bit confused on how it worked, but it's much clearer now.)

After some testing, I also updated the generic to `K extends keyof T & string = keyof T & string`. Before, it was technically possible for `K` to be a non-string value, so I altered this to help enforce the D1 `string` requirement for `colName`.

packages/sdk/src/statement.ts
```diff
@@ -179,6 +180,11 @@ export class Statement<S = unknown> {
     }
   }
 
+  // Check if the first param seen by `first()` is an Options object
+  #checkIsValidOpts(opts: any): opts is Options {
```
dtbuchholz, line 184: More on this below. It does a check on a passed arg to make sure it is a valid `Options` object with a valid `PollingController`. If not, the rest of the logic is handled in the `first` impl.

packages/sdk/src/statement.ts
```diff
@@ -187,30 +193,34 @@ export class Statement<S = unknown> {
...
     opts?: Options
   ): Promise<T[K] | null>;
-  async first<T = S, K extends keyof T = keyof T>(
-    colName?: K,
+  async first<T = S, K extends keyof T & string = keyof T & string>(
+    colNameOrOpts?: K,
     opts: Options = {}
   ): Promise<T | T[K] | null> {
     try {
+      // Handle first overload to ensure passed `opts` are not set to `colName`
```
dtbuchholz, line 211: I discovered that if you try and use the top-most overload for `first()` and pass your own `opts`, the compiler sees the `opts` as `colName`, which obv isn't correct.

for example, running something like this would fail because the passed `opts`/controller was seen as the _first_ param...so then, the `extractColumn` method would try and use the `opts` as the `colName`.
```ts
await stmt.first<{ counter: number; info: string }>({
  controller,
})
```
in other words, before this new new logic, it would see this:
```
colName
{
  controller: {
    signal: AbortSignal { aborted: false },
    abort: [Function: abort],
    interval: 1500,
    cancel: [Function: cancel],
    timeout: 60000
  }
}
opts
{}
```
and with the new logic, it'll properly set colName/opts and use them thereafter.

packages/sdk/src/statement.ts
joewagner: I didn't think about it much.  I ran the tests and saw that there was a type mismatch, then I just copied the D1 types exactly as they are on the [main branch](https://github.com/cloudflare/workerd/blob/main/types/defines/d1.d.ts#L26) and the tests passed.  You probably have more context then me, what was the reason for `T = S[]` (where `S = unknown`)?

joewagner: Looks like it's defined in the constructor, that might work better? Feel free to switch it back.

joewagner (approved):

packages/sdk/src/statement.ts
dtbuchholz: ahh i see what you did re: the tests and 3rd party libs. _nooow_ it makes sense!

and nre: the way in which we implement `raw` with `T = S` seems to work just fine, even though the D1 API expects `T = S[]`. i think it's due to our impl where we return `Array<ValueOf<T>>`, which will properly adhere to the D1 return type `T[]` because of the of `ValueOf`.

sanderpick (approved):
awesome! great work.

PR #65, opened 2023-10-31
Joe/d1 api conformance

joewagner:
@dtbuchholz Here's a few changes to the types to get the third party tests passing without introducing typescript ignores.  If this makes sense to you feel free to do the merge, or manually take some of the changes.

### DIFF
--- a/packages/sdk/src/registry/utils.ts
+++ b/packages/sdk/src/registry/utils.ts
@@ -101,6 +101,7 @@ export function wrapResult<T = unknown>(
     meta,
     success: true,
     results: [],
+    error: undefined,
   };
   if (isTransactionReceipt(resultsOrReceipt)) {
     return { ...result, meta: { ...meta, txn: resultsOrReceipt } };
@@ -159,11 +160,13 @@ export interface Result<T = unknown> {
   /**
    * Whether the query or transaction was successful.
    */
-  success: boolean; // almost always true
+  success: true; // TODO: this is a bug in D1, but if we want to be compatible
+  //       we have to type it like this :<
+  //       https://github.com/cloudflare/workerd/issues/940
   /**
    * If there was an error, this will contain the error string.
    */
-  error?: string; // TODO: this changed to `never` in D1 API
+  error: undefined;
   /**
    * Additional meta information.
    */
--- a/packages/sdk/test/thirdparty.test.ts
+++ b/packages/sdk/test/thirdparty.test.ts
@@ -35,7 +35,8 @@ describe("thirdparty", function () {
   const baseSigner = wallet.connect(provider);
   // Also demonstrates the nonce manager usage
   const signer = new NonceManager(baseSigner);
-  // Name mapping is helpful for Drizzle ORM usage
+  // an `aliases` option is required for Drizzle ORM usage since D1 uses `exec`
+  // internally to do table creation, and we need to capture the uu table name
   const nameMap: NameMapping = {};
   const db = new Database({
     signer,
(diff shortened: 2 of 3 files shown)


packages/sdk/test/thirdparty.test.ts
```diff
@@ -35,7 +35,8 @@ describe("thirdparty", function () {
   const baseSigner = wallet.connect(provider);
   // Also demonstrates the nonce manager usage
   const signer = new NonceManager(baseSigner);
-  // Name mapping is helpful for Drizzle ORM usage
+  // an `aliases` option is required for Drizzle ORM usage since D1 uses `exec`
+  // internally to do table creation, and we need to capture the uu table name
```
joewagner, line 39: Although things might seem to work in some cases with out `aliases`, we should start documenting the `aliases` option as being required to work with the `D1Orm` since any kind of realistic use case will need to have a mapping of table names available to the orm.

packages/sdk/test/thirdparty.test.ts
```diff
@@ -53,7 +54,6 @@ describe("thirdparty", function () {
   });
 
   describe("d1-orm", function () {
-    // @ts-expect-error Tableland database & D1Database type assignment warning
```
joewagner, line 56: The type changes above mean this is no longer needed.  In general, we should never have a ts ignore in our third party tests.

packages/sdk/test/thirdparty.test.ts
```diff
@@ -84,15 +84,9 @@ describe("thirdparty", function () {
...
-      const create = await users.CreateTable({
+      await users.CreateTable({
         strategy: "default",
       });
-      // @ts-expect-error Tableland database & D1Database `ExecResult` type differences
-      await create.txn.wait();
-
-      // TODO: Find a nicer way to deal with this...
-      // @ts-expect-error Tableland database & D1Database `ExecResult` type differences
-      (users.tableName as any) = create.txn.name;
```
joewagner, line 95: If the dev wants to use the D1Orm in a meaningful way they need to use `aliases` and then they never need to think about a blockchain transaction.  
There's a tricky point here though: with chains that take a long time to do creates, devs might not want to use the ORM for the table creation.  If the block takes too long, or the process running the create loses network connectivity, or anything happens to before the alias file is written to it might seem like the table wasn't created when it actually was.  Hopefully we can explain this with docs and a playbook.

packages/sdk/src/registry/utils.ts
```diff
@@ -159,11 +160,13 @@ export interface Result<T = unknown> {
   /**
    * Whether the query or transaction was successful.
    */
-  success: boolean; // almost always true
+  success: true; // TODO: this is a bug in D1, but if we want to be compatible
+  //       we have to type it like this :<
+  //       https://github.com/cloudflare/workerd/issues/940
```
joewagner, line 165: Note that this _should_ probably still be `boolean` but D1 changed the type so we have to do the same.

packages/sdk/test/thirdparty.test.ts
dtbuchholz: makes sense, created an [issue in docs](https://linear.app/tableland/issue/DOC-90/updateadd-orm-docs-and-requirement-to-use-the-sdk-aliases-feature)

packages/sdk/test/thirdparty.test.ts
dtbuchholz: cool, makes sense—add to as an [issue in docs](https://linear.app/tableland/issue/DOC-90/updateadd-orm-docs-and-requirement-to-use-the-sdk-aliases-feature), along with the context below

PR #67, opened 2023-11-04
feat: add `node-helpers` package & refactor for major bump

dtbuchholz:
## Summary

Adds a `node-helpers` package for `jsonFileAliases`. Also, a couple of small cleanups are included for the SDK: removing deprecated `createTable` and `RunSQLParams` as well as small package.json updates (GH repo URL, misspelled license file).

## Details

The bulk of the changes include a refactor to pull `jsonFileAliases` into a `node-helpers` package; the CLI and SDK were updated accordingly with usage/tests.

## How it was tested

- Added tests for `node-helpers`.
- Updated CLI & SDK tests that previously were using their own `jsonFileAliases` methods, but now it's an import.

## Checklist:


packages/cli/src/lib/commandSetup.ts
```diff
@@ -1,8 +1,9 @@
 import { helpers, Database, Registry, Validator } from "@tableland/sdk";
 import { init } from "@tableland/sqlparser";
+import { jsonFileAliases } from "@tableland/node-helpers";
```
dtbuchholz, line 3: CLI now uses the `node-helpers` package and removes the duplicative definitions for `jsonFileAliases`—the tests reflect these changes as well (a _slightly_ different error message).

packages/cli/src/utils.ts
```diff
@@ -1,16 +1,13 @@
...
+import { jsonFileAliases } from "@tableland/node-helpers";
 
-// TODO: should be able to remove the "staging" filter since the SDK handles it
 export const getChains = function (): typeof helpers.supportedChains {
-  return Object.fromEntries(
-    Object.entries(helpers.supportedChains).filter(
-      ([name]) => !name.includes("staging")
-    )
-  ) as Record<helpers.ChainName, helpers.ChainInfo>;
+  return helpers.supportedChains;
```
dtbuchholz, line 8: don't need this filtering anymore now that the SDK does this for us

packages/cli/test/create.test.ts
```diff
@@ -167,7 +168,7 @@ describe("commands/create", function () {
     const privateKey = account.privateKey.slice(2);
     const consoleError = spy(logger, "error");
     // Set up faux aliases file
-    const aliasesFilePath = "./invalid.json";
+    const aliasesFilePath = "./path/to/invalid.json";
```
dtbuchholz, line 171: the way i originally implemented the CLI's `jsonFileAliases` didn't let you define a non-existent file at a valid path. (i think that differed from the SDK's implementation.) the tests were changed to try and pass an invalid path/file in order to properly fail.

packages/node-helpers/README.md
```diff
@@ -0,0 +1,11 @@
+# `@tableland/node-helpers`
```
dtbuchholz, line 1: (will update this...)

packages/node-helpers/src/utils.ts
```diff
@@ -0,0 +1,83 @@
+import { readFileSync, statSync, writeFileSync, existsSync } from "node:fs";
+import { extname, resolve, dirname } from "node:path";
+
+/**
+ * A series of mappings from a table alias to its globally unique table name.
+ */
+export type NameMapping = Record<string, string>;
```
dtbuchholz, line 7: the one callout—in order to avoid nx circular dependencies, i have to define these types here vs. importing them from the SDK. it might be useful in the future to have a `@tableland/types` package to help with this.

packages/node-helpers/src/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+/**
+ * Get the path to where an aliases file exists or should be created.
+ * @param path Path to an existing aliases file, a directory where one
+ * should be created (with a default `tableland.aliases.json` name), or a
+ * directory + the desired filename that should be created.
+ * @returns The full path of the aliases file.
+ */
+function getFilePath(path: string): string {
```
dtbuchholz, line 24: i created this new helper so that in `jsonFileAliases`, we:
1) check the path is valid
2) if so, check if it's a filepath or directory path:
   - if it's a file, then we return the path as long as the file is valid `.json`
   - if it's a dir, then we use a default filename `tableland.aliases.json` and create/use that file for them later
3) if the path does not exist, we check if the _directory_ exists—if so, that means the user passed a valid path but non-existent filename, so we should create the file for them using their custom filename
4) if all else fails, that means the path does not exist, or the path provided a file that wasn't valid json

packages/node-helpers/src/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+ * and its `Database` class.
+ * @param path The path to an existing aliases file, a directory + filename
+ * path where one should be created, or a directory path where a default
+ * `tableland.aliases.json` filename should be created.
+ * @returns An {@link AliasesNameMap} for table alias functionality.
+ */
+export function jsonFileAliases(path: string): AliasesNameMap {
+  // Ensure the path exits & retrieve it: either a file, a directory, or a
+  // directory that defines a non-existent filename (to be created).
+  const filepath = getFilePath(path);
```
dtbuchholz, line 69: here's where we make the check to get the proper path, based on whatever the user provided. the main rationale here is that it:
- ensures the path exists
- if a directory was provided, that means `jsonFileAliases` was created a with a dir path, so we need to make sure the default filename `tableland.aliases.json` is appended properly

packages/sdk/src/registry/run.ts
```diff
@@ -26,23 +26,12 @@ export async function prepareMutateOne({
   statement,
   chainId,
   first,
-}: PrepareParams): Promise<RunSQLParams & { prefix: string }> {
+}: PrepareParams): Promise<MutateOneParams & { prefix: string }> {
```
dtbuchholz, line 29: we dropped `RunSQLParams` (deprecated), so now we just use `MutateOneParams`

packages/sdk/test/registry.test.ts
```diff
@@ -32,7 +32,7 @@ describe("registry", function () {
   describe("controller", function () {
     let receipt: MultiEventTransactionReceipt;
     this.beforeAll(async function () {
-      const tx = await reg.createTable({
+      const tx = await reg.create({
```
dtbuchholz, line 35: we dropped `createTable` (deprecated), so now we just use `create`

packages/sdk/src/helpers/config.ts
```diff
@@ -82,43 +88,6 @@ export async function extractChainId(conn: Config = {}): Promise<number> {
   return chainId;
 }
 
-const findOrCreateFile = async function (filepath: string): Promise<Buffer> {
```
dtbuchholz, line 85: we dropped this SDK helper as well as `jsonFileAliases` since it's now part of `node-helpers`

dtbuchholz:
one random question not related to this PR. 

even before this PR, i noticed that when i run the tests locally, the following CLI test _always_ fails. but, it succeeds when i push to GH, so it's not an "issue" per-se. idk why it's failing on my machine because the error is definitely `cannot determine provider chain ID` if i run the actual command...maybe it's some problem related to local tableland processes?
```
@tableland/cli:   1) commands/create
@tableland/cli:        create can accept custom providerUrl:
@tableland/cli:       AssertionError [ERR_ASSERTION]: 'This operation was aborted' == 'cannot determine provider chain ID'
@tableland/cli:       + expected - actual
@tableland/cli:       -This operation was aborted
@tableland/cli:       +cannot determine provider chain ID
@tableland/cli:       
@tableland/cli:       at Context.<anonymous> (file:///Users/dtb/tbl/tableland-js/packages/cli/test/create.test.ts:442:5)
@tableland/cli:       at processTicksAndRejections (node:internal/process/task_queues:95:5)
@tableland/cli: npm ERR! Lifecycle script `coverage` failed with error: 
@tableland/cli: npm ERR! Error: command failed 
@tableland/cli: npm ERR!   in workspace: @tableland/cli@5.3.0 
@tableland/cli: npm ERR!   at location: /Users/dtb/tbl/tableland-js/packages/cli 
@tableland/cli: npm ERR! Lifecycle script `test:ci` failed with error: 
@tableland/cli: npm ERR! Error: command failed 
@tableland/cli: npm ERR!   in workspace: @tableland/cli@5.3.0 
@tableland/cli: npm ERR!   at location: /Users/dtb/tbl/tableland-js/packages/cli 
```
this is the test:
```ts
  test("create can accept custom providerUrl", async function () {
    const [account] = accounts;
    const privateKey = account.privateKey.slice(2);
    const consoleError = spy(logger, "error");

    await yargs([
      "create",
      "id int primary key, name text",
      "--prefix",
      "custom_url_table",
      ...defaultArgs,
      "--privateKey",
      privateKey,
      "--providerUrl",
      "http://localhost:9876",
    ])
      .command(mod)
      .parse();

    const value = consoleError.getCall(0).firstArg;
    equal(value, "cannot determine provider chain ID");
  });
```

packages/node-helpers/src/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+  throw new Error(`invalid aliases path`);
+}
+
+/**
+ * Either find an existing table aliases file or create one.
+ * @param path Path to an existing aliases file, or the directory where one
+ * should be created (can be a directory path, or directory + filename path).
+ * @returns A buffer containing the JSON file contents.
+ */
+const findOrCreateFile = async function (path: string): Promise<Buffer> {
```
joewagner, line 49: Can this be a sync function?

joewagner (dismissed):
Looks good @dtbuchholz!
Not exactly related to your changes, but after looking through this I'm wondering if the `AliasesNameMap` interface should allow `read` and `write` to be sync or async? 
If that makes sense, it would be nice to make that change as part of the new major version.

joewagner:
> one random question not related to this PR.
> 
> even before this PR, i noticed that when i run the tests locally, the following CLI test _always_ fails. but, it succeeds when i push to GH, so it's not an "issue" per-se. idk why it's failing on my machine because the error is definitely `cannot determine provider chain ID` if i run the actual command...maybe it's some problem related to local tableland processes?
> 
> ```
> @tableland/cli:   1) commands/create
> @tableland/cli:        create can accept custom providerUrl:
> @tableland/cli:       AssertionError [ERR_ASSERTION]: 'This operation was aborted' == 'cannot determine provider chain ID'
> @tableland/cli:       + expected - actual
> @tableland/cli:       -This operation was aborted
> @tableland/cli:       +cannot determine provider chain ID
> @tableland/cli:       
> @tableland/cli:       at Context.<anonymous> (file:///Users/dtb/tbl/tableland-js/packages/cli/test/create.test.ts:442:5)
> @tableland/cli:       at processTicksAndRejections (node:internal/process/task_queues:95:5)
> @tableland/cli: npm ERR! Lifecycle script `coverage` failed with error: 
> @tableland/cli: npm ERR! Error: command failed 
> @tableland/cli: npm ERR!   in workspace: @tableland/cli@5.3.0 
> @tableland/cli: npm ERR!   at location: /Users/dtb/tbl/tableland-js/packages/cli 
> @tableland/cli: npm ERR! Lifecycle script `test:ci` failed with error: 
> @tableland/cli: npm ERR! Error: command failed 
> @tableland/cli: npm ERR!   in workspace: @tableland/cli@5.3.0 
> @tableland/cli: npm ERR!   at location: /Users/dtb/tbl/tableland-js/packages/cli 
> ```
> 
> this is the test:
> 
> ```ts
>   test("create can accept custom providerUrl", async function () {
>     const [account] = accounts;
>     const privateKey = account.privateKey.slice(2);
>     const consoleError = spy(logger, "error");
> 
>     await yargs([
>       "create",
>       "id int primary key, name text",
>       "--prefix",
>       "custom_url_table",
>       ...defaultArgs,
>       "--privateKey",
>       privateKey,
>       "--providerUrl",
>       "http://localhost:9876",
>     ])
>       .command(mod)
>       .parse();
> 
>     const value = consoleError.getCall(0).firstArg;
>     equal(value, "cannot determine provider chain ID");
>   });
> ```

Try changing `localhost` to `127.0.0.1` and see if it passes for you?

dtbuchholz:
> Not exactly related to your changes, but after looking through this I'm wondering if the AliasesNameMap interface should allow read and write to be sync or async?
If that makes sense, it would be nice to make that change as part of the new major version.

@joewagner sure, will do that. and i figured out the test issue—it was using the `defaultArgs`, which already defines a `--providerUrl`. if i remove that and also pass the `--chain local-tableland`, it fixes the test.

dtbuchholz:
Okay, I've refactored everything so methods and types are split into synchronous vs. asynchronous file operations. The `AliasesNameMap` unifies the sync vs. async types together, and the `jsonFileAliases` uses an overload to determine which type of behavior should be used. 

Everything should still work in the CLI/SDK without any changes. The usage allows for an optional `async` flag to be passed and looks like:
```js
const aliasesSync = jsonFileAliases("/path/to/file") // default to synchronous behavior
const aliasesAsync = await jsonFileAliases("/path/to/file", true) // use asynchronous behavior
```

But, lmk if this wasn't the right approach!

packages/node-helpers/src/utils.ts
```diff
@@ -0,0 +1,318 @@
...
+export interface AliasesNameMapAsync {
+  read: () => Promise<NameMapping>;
+  write: (map: NameMapping) => Promise<void>;
+}
+
+/**
+ * Used to read and write table aliases within a `Database` instance. See
+ * {@link AliasesNameMapAsync} and {@link AliasesNameMapSync} for more details.
+ */
+export type AliasesNameMap = AliasesNameMapSync | AliasesNameMapAsync;
```
dtbuchholz, line 46: here, i'm allowing `AliasesNameMap` to be either a sync or async variation, which is what the Database class is ultimately using, too.

packages/node-helpers/src/utils.ts
```diff
@@ -0,0 +1,318 @@
...
+
+/**
+ * Validate that a path exists and is readable, synchronously.
+ * @param path Path to a file or directory.
+ * @returns `true` if the path exists and is readable, `false` otherwise.
+ */
+function canReadPathSync(path: string): boolean {
```
dtbuchholz, line 53: i created some helpers for `canRead` or `canWrite`. since async operations don't have node method like `existsSync`, i had to use `access`. thus, i figured it'd be clearer if all of the sync methods followed a similar pattern with `accessSync` to check read/write access on a provided file path.

packages/node-helpers/src/utils.ts
```diff
@@ -0,0 +1,318 @@
...
+  }
+}
+
+/**
+ * Validate that a path exists and is both readable & writable, synchronously.
+ * @param path Path to a file or directory.
+ * @returns `true` if the path exists and is both readable & writable, `false`
+ * otherwise.
+ */
+const pathExistsSync = (path: string): boolean =>
```
dtbuchholz, line 110: these basically do the same thing as `existsSync` but with sync/async variations, and the approach is _bit_ better, too, since it ensures the path can be read/written to.

packages/node-helpers/src/utils.ts
```diff
@@ -0,0 +1,318 @@
...
+
+/**
+ * Get the path to where an aliases file exists or should be created,
+ * synchronously.
+ * @param path Path to an existing aliases file, a directory where one
+ * should be created (with a default `tableland.aliases.json` name), or a
+ * directory + the desired filename that should be created.
+ * @returns The full path of the aliases file.
+ */
+function getFilePathSync(path: string): string {
```
dtbuchholz, line 130: sync/async variations for `getFilePath` to check a file exists and can be read/written to

packages/node-helpers/src/utils.ts
```diff
@@ -0,0 +1,318 @@
...
+/**
+ * Either find an existing table aliases file or create one, with an option for
+ * synchronous (default) or asynchronous behavior.
+ * @param path Path to an existing aliases file, or the directory where one
+ * should be created (can be a directory path, or directory + filename path).
+ * @param async Whether to use asynchronous or synchronous file operations.
+ * Defaults to `false` for synchronous behavior.
+ * @returns A buffer containing the JSON file contents.
+ */
+export function findOrCreateAliasesFile(
```
dtbuchholz, line 217: this gets used in the CLI (i removed duplicative logic that i had implemented there in the past). it follows similar logic to `jsonFileAliases` where the default is synchronous behavior, but an async option can be set.

packages/node-helpers/src/utils.ts
```diff
@@ -0,0 +1,318 @@
...
+export function jsonFileAliases(path: string): AliasesNameMapSync;
+export function jsonFileAliases(
+  path: string,
+  async?: false
+): AliasesNameMapSync;
+export function jsonFileAliases(
+  path: string,
+  async?: true
+): Promise<AliasesNameMapAsync>;
+export function jsonFileAliases(
```
dtbuchholz, line 309: here's where we overload `jsonFileAliases` to allow the same method to be used for sync/async variations, set by the `async` option. i just wanna double check this is the correct approach vs. splitting into separate methods like `jsonFileAliasesSync` and `jsonFileAliasesAsync`.

joewagner:
> Okay, I've refactored everything so methods and types are split into synchronous vs. asynchronous file operations. The `AliasesNameMap` unifies the sync vs. async types together, and the `jsonFileAliases` uses an overload to determine which type of behavior should be used. See [here](https://github.com/tablelandnetwork/tableland-js/pull/67/files/971336165ed1b1c827adee90f95b3e00d4f3161f#diff-8bb588f1a0a2876e7d60f66bf88d7537f014209b1db47de92d760f35d5068d75) for the most recent changes.

Sorry I should have explained a bit better what I was thinking.  I was trying to point out that `jsonFileAliases` was **not** using any calls to await, and you had made it an `async` function.  I think leaving it as a single "non-async" function makes sense, but the SDK api would need to update the `AliasesNameMap` interface to something like:
```
interface AliasesNameMap {
  read: () => Promise<NameMapping> | NameMapping;
  write: (map: NameMapping) => Promise<void> | void;
}
```
Then check what kind of function the user had passed in when using the name mapper.

I should have just opened a PR at first.  Here is what I was thinking: #68

dtbuchholz:
@joewagner ahh, i see...yeah that makes way more sense, no worries! and thanks for the cleanups, i can merge your branch here if that works for you...it's marked as a draft PR, so i just want to double check before i do anything here.

joewagner:
> @joewagner ahh, i see...yeah that makes way more sense, no worries! and thanks for the cleanups, ~i can merge your branch here if that works for you...it's marked as a draft PR, so i just want to double check before i do anything here.~
> 
> edit: i merged those changes in here. also, with the `@tableland/sdk` as a dev dep, are you basically saying that we can kinda "ignore" the warning since we know why it's happening? e.g., i still get this when `node-helpers` uses the SDK as a dev dep:

I had not noticed that warning.  Looking at everything again, and it seems like we move the tests for the helper -> sdk integration out of the sdk into the helper.  I think that makes the most sense, since users of the helper will be importing the SDK and the helper package.


> as for the following:
> 
> > Then check what kind of function the user had passed in when using the name mapper.
> 
> i can take a stab at this. if i understand correctly, the SDK's `Statement`, `Database`, etc are doing a check like this:
> 
> ```ts
> const nameMap =
>   typeof this.config.aliases?.read === "function"
>     ? await this.config.aliases.read()
>     : undefined;
> ```
> 
> this would assume that `read()` is always returning a promise when extracting the `NameMapping`, which isn't the case for something like `jsonFileAliases`. so, i can alter this with a helper so that this snippet now looks like this:
> 
> ```ts
> const nameMap =
>   typeof this.config.aliases?.read === "function"
>     ? await readNameMapping(this.config.aliases)
>     : undefined;
> ```
> 
> where helpers can be used to determine if it's a sync or async `AliasesNameMap.read()` method, like:
> 
> ```ts
> export async function readNameMapping(
>   aliases: AliasesNameMap
> ): Promise<NameMapping> {
>   const nameMap = aliases.read();
>   return isPromise(nameMap) ? await nameMap : nameMap;
> }
> 
> // helper
> export function isPromise(arg: any): arg is Promise<any> {
>   return Boolean(
>     arg != null && typeof arg === "object" && typeof arg.then === "function"
>   );
> }
> ```

Yep thats the idea!  I'll push the test changes now, if you want to make the changes to accessing the different types of alias functions

dtbuchholz:
@joewagner re: tests—got it, makes sense to me. and i've added the `readNameMapping ` helper as well as one for `writeNameMapping`.

joewagner (approved):
Looks great, nice work on all of this!

PR #81, opened 2023-11-14
[chore] Bump @tableland/evm

avichalp:
## Summary

Updates the version of tableland/evm package which supports Arbitrum sepolia instead of arbitrum goerli

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #82, opened 2023-11-15
[chore] Bump sdk version in cli and local

avichalp:
## Summary

Updates the version of tableland/sdk package which uses the latest tableland/evm

## Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Any dependent changes have been merged and published in downstream modules


packages/local/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tableland/local",
-  "version": "2.2.0",
+  "version": "2.3.0",
```
joewagner, line 3: I think this value was already updated in #81.  I think we want this to be 2.2.0

packages/cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tableland/cli",
-  "version": "5.3.2",
+  "version": "5.4.2",
```
joewagner, line 3: The next version here would be 5.4.0 right?

avichalp: oh yeah! updating

packages/local/package.json
joewagner: Sorry, this is a bit confusing.  The version of local has already been updated in #81 here:
https://github.com/tablelandnetwork/tableland-js/pull/81/files#diff-eebd2b35696842ddf3f1a4f7248077fb077628ca056a247012c7b24ca646254dR3

You can see that the current version pulbished to npm is 2.1.1 
https://www.npmjs.com/package/@tableland/local?activeTab=versions

Which means the next version to be published should be 2.2.0 right?

joewagner (approved):

packages/local/package.json
avichalp: ah... got it! the version was updated in #81 but not published. thanks for the explanation, fixed it.

PR #84, opened 2023-11-20
Enable usage with Node.js 20 LTS version

joewagner:
## Overview 
This upgrades the `@tableland/local` `hardhat` dependency version, and switches our invalid JSON error message detection to something that works for all LTS versions of node.js

### CI
add node version 20 to the matrix

### SDK
Tests were updated to more lenient error message matching

### Local
`local`'s current version of Hardhat does not work with Node.js version 20.x.x  
If you try to use it you will receive the following error
```
unrecoverable error
Error: WARNING: You are using a version of Node.js that is not supported, and it may work incorrectly, or not work at all. See https://hardhat.org/nodejs-versions
```
Note, the newest version of `hardhat` works with all node.js LTS versions

### CLI
Only a small change needed here.  The read command is using string matching to detect invalid json as a result of incorrect use of the `--wrap` flag.  We have to update that string matching.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,17 +14,27 @@
     "build": "npx lerna run build"
   },
   "devDependencies": {
+    "@databases/escape-identifier": "^1.0.3",
+    "@databases/sql": "^3.2.0",
+    "@ethersproject/experimental": "^5.7.0",
+    "@playwright/test": "^1.30.0",
     "@types/chai": "^4.3.5",
     "@types/chai-as-promised": "^7.1.5",
+    "@types/cosmiconfig": "^6.0.0",
     "@types/cross-spawn": "^6.0.2",
     "@types/inquirer": "^9.0.3",
+    "@types/js-yaml": "^4.0.5",
     "@types/mocha": "^10.0.1",
     "@types/shelljs": "^0.8.12",
+    "@types/sinon": "^10.0.20",
+    "@types/yargs": "^17.0.19",
     "@typescript-eslint/eslint-plugin": "^5.62.0",
     "@typescript-eslint/parser": "^5.62.0",
     "c8": "^8.0.1",
     "chai": "^4.3.7",
     "chai-as-promised": "^7.1.1",
+    "d1-orm": "^0.9.0",
+    "drizzle-orm": "^0.29.0",
     "eslint": "^8.32.0",
     "eslint-config-prettier": "^8.5.0",
     "eslint-config-standard": "^17.0.0",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -14,17 +14,27 @@
     "build": "npx lerna run build"
   },
   "devDependencies": {
+    "@databases/escape-identifier": "^1.0.3",
+    "@databases/sql": "^3.2.0",
+    "@ethersproject/experimental": "^5.7.0",
+    "@playwright/test": "^1.30.0",
```
joewagner, line 20: devDependencies were being listed in each package, best practice afaik is to put them all in the top level

packages/cli/src/commands/read.ts
```diff
@@ -101,7 +101,7 @@ export const handler = async (argv: Arguments<Options>): Promise<void> => {
           unwrap: argv.unwrap,
         });
       } catch (err: any) {
-        if (err.message.includes("in JSON at position") as boolean) {
+        if (err.message.includes("JSON at position") as boolean) {
```
joewagner, line 104: Aside from upgrading `hardhat` the big change is the error message returned when invalid json is parsed has change slightly.

packages/sdk/package.json
```diff
@@ -53,11 +53,6 @@
     "dist/**/package.json",
     "src/**/*.ts"
   ],
-  "nx": {
-    "implicitDependencies": [
-      "!@tableland/local"
```
joewagner, line 58: the most recent `nx` removed that need for this.

dtbuchholz (approved):

dtbuchholz:
@joewagner a random question re: SDK compatibility. in the past, i remember some issues with the Node `fetch` API such that we said devs needed to use Node 18+ or set up a polyfill.

is that still the case? or is it more accurate to state Node 16+ is supported?

also, if anything comes to mind on [compatibility](https://docs.tableland.xyz/sdk/reference/compatability) callouts, lmk. i've kept that docs page pretty basic, but i'd imagine things like webpack, the package.json ESx target/module versions, etc. might be nice to include.

joewagner:
> @joewagner a random question re: SDK compatibility. in the past, i remember some issues with the Node `fetch` API such that we said devs needed to use Node 18+ or set up a polyfill.
> 
> is that still the case? or is it more accurate to state Node 16+ is supported?

Node 16 _should_ work if you include the [--experimental-fetch flag](https://nodejs.org/dist/latest-v16.x/docs/api/cli.html#--experimental-fetch).  Although folks might want to use a polyfill depending on what they are trying to do, and in general we should encourage everyone to use the newest LTS version.

> also, if anything comes to mind on [compatibility](https://docs.tableland.xyz/sdk/reference/compatability) callouts, lmk. i've kept that docs page pretty basic, but i'd imagine things like webpack, the package.json ESx target/module versions, etc. might be nice to include.

Will do, what you have currently looks good to me.

PR #99, opened 2023-11-21
fix: mumbai failing to due gas limit issues

### LINKED ISSUE #98: [TABJS-19] Fix polygon mumbai gas limit issue causing failed txs
## Summary

When creating or writing to a table on `maticmum` with the SDK or CLI, you run into an `UNPREDICTABLE_GAS_LIMIT` error. This is due to the gas limit being set about 20% lower than it should be, and it only affect polygon.

## Details

See [this failed tx](https://dashboard.tenderly.co/tx/80001/0x175ad3994cf2e67dbd9f61c27d7075fec84ba0023406d622498884b04c0e5fff) for more details and the ethers error below:

```
cannot estimate gas; transaction may fail or may require manual gas limit [ See: https://links.ethers.org/v5-errors-UNPREDICTABLE_GAS_LIMIT ]
```

It was a simple 1 col insert with gas used 51,416, and gas limit 51,757. If you make this tx on polygonscan + MM, the gas limit is around 70-80k.

### Solution

A quick fix looks something like this but where `gasLimit` is then bumped by 20% for polygon only: [https://gist.github.com/dtbuchholz/b24174ecb3e8b6923a2b2a30e9aa286f](https://gist.github.com/dtbuchholz/b24174ecb3e8b6923a2b2a30e9aa286f)

<sub>From [SyncLinear.com](https://synclinear.com) | [TABJS-19](https://linear.app/tableland/issue/TABJS-19/fix-polygon-mumbai-gas-limit-issue-causing-failed-txs)</sub>

dtbuchholz:
## Summary

This fixes an issue where Mumbai txs were always failing for both creates and writes. It bumps the gas limit up by 20% for all Polygon chains. The CLI package has been updated with a patch bump to incorporate the new SDK version's patch. Also, small package.json updates were made for GH repo links in all packages. Closes #98.

## Details

It affects all state altering methods. This includes `create` and `mutate` methods for both single and batch statements, setting/locking controllers, and transfers.

## How it was tested

You can see a test create and write tx here, where the gas limit leaves about 20% of extra room:
- create: [0xd36a73256b38b866ebda65dff619174145768f35a2cb179f2ebc0f3832344bbc](https://mumbai.polygonscan.com/tx/0xd36a73256b38b866ebda65dff619174145768f35a2cb179f2ebc0f3832344bbc)
- mutate: [0x236fa88bb9034629f9f4ea8e23a28ddd32bc6f13bb11ff1c602051e4689adf92](https://mumbai.polygonscan.com/tx/0x236fa88bb9034629f9f4ea8e23a28ddd32bc6f13bb11ff1c602051e4689adf92)

Prior to this, it'd fail with a 99% gas usage, like [here](https://mumbai.polygonscan.com/tx/0xfff0756b73d96098e6bd7a8edb0cbf4a0d24071d79609b0330331a5885dd6814).

The controller and transfer methods don't _necessarily_ need this logic and pass without needing to introduce the gas limit bump. These are passing transactions from the existing logic, before introducing the 20% bump:
- controller: [0x685a7b9e0b67d19c06be97246e88672d0604777699e5816b8da44248377a237f](https://mumbai.polygonscan.com/tx/0x685a7b9e0b67d19c06be97246e88672d0604777699e5816b8da44248377a237f)
- transfer: [0x4294f15e9704492d8e86c39e245137b306a7acf59e88be79e1be1541efcb6df7](https://mumbai.polygonscan.com/tx/0x4294f15e9704492d8e86c39e245137b306a7acf59e88be79e1be1541efcb6df7)

However, you can see the transfer was at 99%, and controller was at 87% of the gas limit. Thus, it makes sense to also implement the gas limit bump for them as a margin of safety. Here are changes with the 20% bump included:
- controller: [0x16ae25088b56b6703e51bfbd8fff781f3dd6f5f47988c9171e3eddf021762029](https://mumbai.polygonscan.com/tx/0x16ae25088b56b6703e51bfbd8fff781f3dd6f5f47988c9171e3eddf021762029)
- transfer: [0x487391468cf78f0702bdb4e85c5bc625cd94d7fede542135148c2d586a36ec84](https://mumbai.polygonscan.com/tx/0x487391468cf78f0702bdb4e85c5bc625cd94d7fede542135148c2d586a36ec84)

## Checklist:

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,8 +1,12 @@
 {
   "name": "@tableland/cli",
-  "version": "5.4.0",
+  "version": "5.4.1",
   "description": "Tableland command line tools",
-  "repository": "https://github.com/tablelandnetwork/tableland-js",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/tablelandnetwork/tableland-js.git",
+    "directory": "packages/cli"
+  },
   "publishConfig": {
     "access": "public"
   },
@@ -46,7 +50,7 @@
   "dependencies": {
     "@ensdomains/ensjs": "3.0.0-alpha.64",
     "@tableland/node-helpers": "^0.0.2",
-    "@tableland/sdk": "^5.1.0",
+    "@tableland/sdk": "^5.1.1",
     "@tableland/sqlparser": "^1.3.0",
     "cli-select-2": "^2.0.0",
     "cosmiconfig": "^8.0.0",
--- a/packages/local/package.json
+++ b/packages/local/package.json
@@ -2,7 +2,11 @@
   "name": "@tableland/local",
   "version": "2.3.0",
   "description": "Tooling to start a sandboxed Tableland network.",
-  "repository": "https://github.com/tablelandnetwork/tableland-js",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/tablelandnetwork/tableland-js.git",
+    "directory": "packages/local"
+  },
   "license": "MIT AND Apache-2.0",
   "publishConfig": {
     "access": "public"
--- a/packages/node-helpers/package.json
+++ b/packages/node-helpers/package.json
@@ -2,7 +2,11 @@
   "name": "@tableland/node-helpers",
   "version": "0.0.2",
   "description": "Helpers for the @tableland/sdk in a Node.js environment",
-  "repository": "https://github.com/tablelandnetwork/tableland-js",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/tablelandnetwork/tableland-js.git",
+    "directory": "packages/node-helpers"
+  },
   "license": "MIT AND Apache-2.0",
   "publishConfig": {
     "access": "public"
--- a/packages/sdk/src/helpers/ethers.ts
+++ b/packages/sdk/src/helpers/ethers.ts
@@ -29,7 +29,7 @@ export async function getOverrides({
   const opts: Overrides = {};
   const network = await signer.provider?.getNetwork();
   /* c8 ignore next 7 */
-  if (network?.chainId === 137) {
+  if (network && isPolygon(network.chainId)) {
     const feeData = await signer.getFeeData();
     if (feeData.gasPrice != null) {
       opts.gasPrice =
@@ -37,6 +37,15 @@ export async function getOverrides({
     }
   }
   return opts;
+}
+
+/**
+ * Check if a chain is Polygon.
+ * @param chainId The chainId of the network.
+ * @returns A boolean that indicates if the chain is a mainnet/testnet Polygon.
+ */
+export function isPolygon(chainId: number): boolean {
+  return chainId === 137 || chainId === 80001;
 }
 
 /**
--- a/packages/sdk/src/registry/create.ts
+++ b/packages/sdk/src/registry/create.ts
@@ -1,6 +1,6 @@
 import { normalize } from "../helpers/index.js";
 import { type SignerConfig } from "../helpers/config.js";
-import { type ContractTransaction } from "../helpers/ethers.js";
+import { type ContractTransaction, isPolygon } from "../helpers/ethers.js";
 import { validateTableName } from "../helpers/parser.js";
 import { getContractAndOverrides } from "./contract.js";
 
@@ -96,6 +96,14 @@ async function _createOne(
     signer,
     chainId
   );
+  if (isPolygon(chainId)) {
+    const gasLimit = await contract.estimateGas["create(address,string)"](
+      owner,
+      statement,
+      overrides
+    );
+    overrides.gasLimit = Math.floor(gasLimit.toNumber() * 1.2);
+  }
   return await contract["create(address,string)"](owner, statement, overrides);
 }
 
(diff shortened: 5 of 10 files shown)


packages/sdk/src/helpers/ethers.ts
```diff
@@ -39,6 +39,15 @@ export async function getOverrides({
   return opts;
 }
 
+/**
+ * Check if a chain is Polygon.
+ * @param chainId The chainId of the network.
+ * @returns A boolean that indicates if the chain is a mainnet/testnet Polygon.
+ */
+export function isPolygon(chainId: number): boolean {
```
dtbuchholz, line 47: made this a helper so it can be used in the registry methods, too. also, it was only including mainnet polygon, not testnet, so that's been added

packages/sdk/src/registry/create.ts
```diff
@@ -96,6 +96,14 @@ async function _createOne(
     signer,
     chainId
   );
+  if (isPolygon(chainId)) {
```
dtbuchholz, line 99: all of the state altering methods implement this pattern to estimate gas for a method call. 

I debated making this as part of `getContractAndOverrides` to reduce somewhat duplicative code—like passing the fn signature `"create(address,string)"` and other params there. but it wasnt worth the effort imo.

dtbuchholz:
@joewagner question on the package version bumps. since everything uses the `@tableland/sdk`, is it a best practice to _always_ publish new versions for cli/local/node-helpers if the SDK gets a new release?

i know the CLI always needs a bump to incorporate SDK changes. but something like node-helpers only uses the SDK as a dev dep. i figured it's ideal to keep things in sync but wanted to double check.

joewagner:
> @joewagner question on the package version bumps. since everything uses the `@tableland/sdk`, is it a best practice to _always_ publish new versions for cli/local/node-helpers if the SDK gets a new release?

Not necessarily.  The basic rule is, if you want a new version of a package to be published on npm you will need to update the version for that package.

Here are two examples that demonstrate some more complicated scenarios:
1. We created a new feature in the sdk and want that to be available on npm. In this case we would increment the sdk minor version.  `cli` and `local` don't have anything new to be published on npm, so they are not changed.
2. There is a fix for a bug in the sdk that makes it's way downstream and causes a bug in `local` but **not** in `cli`.  In this example we would increment the `sdk` and `local` patch versions.

Nuances of these two examples:
- In example 1 the `dependencies` in `local` and `cli` don't need any change since they are `^x.x.x` which means that users of our packages can have the flexibility to use the combinations of versions they want to use
- In example 2 the `local` `dependencies` entry for `@tableland/sdk` should to be updated to whatever the new `sdk` version is... Obviously still using the `^` syntax.

> i know the CLI always needs a bump to incorporate SDK changes. but something like node-helpers only uses the SDK as a dev dep. i figured it's ideal to keep things in sync but wanted to double check since node-helpers was previously using 5.0.0 whereas the CLI/local were using 5.1.0. wasnt sure if that was intentional or not.

The CLI does not always need a bump, but if you want to publish a new version of the CLI (because of a bug fix, or added feature) you **do** need to bump the version.

As an example of a time bumping a version would not make sense, imagine there is a bug in the sdk's implementation of `exec`.  The CLI never uses `exec`, and there isn't any need to publish a new version of it.

It's all potentially a little complicated, does that answer your question?

packages/local/package.json
```diff
@@ -60,7 +64,7 @@
     "up:dev": "node dist/esm/up.js --validator ../go-tableland --registry ../evm-tableland"
   },
   "dependencies": {
-    "@tableland/sdk": "^5.1.0",
+    "@tableland/sdk": "^5.1.1",
```
joewagner, line 67: Since this package doesn't ever interact with mumbai is there a reason to republish it, and require users of it to upgrade their version of the sdk?

packages/node-helpers/package.json
```diff
@@ -43,7 +47,7 @@
     "build": "npm run build:esm && npm run build:cjs && ./fixup"
   },
   "devDependencies": {
-    "@tableland/sdk": "^5.0.0",
+    "@tableland/sdk": "^5.1.1",
```
joewagner, line 50: Same question as above.  Nothing in this package has been modified, so we might as well remain flexible in what versions consumers can have as deps and peer deps.

packages/cli/package.json
```diff
@@ -46,7 +50,7 @@
   "dependencies": {
     "@ensdomains/ensjs": "3.0.0-alpha.64",
     "@tableland/node-helpers": "^0.0.2",
-    "@tableland/sdk": "^5.1.0",
+    "@tableland/sdk": "^5.1.1",
```
joewagner, line 53: I'm thinking this does need to be bumped since the polygon bug affects users of the cli right?
As such, the above new patch version for the cli also make sense.

joewagner (commented):
@dtbuchholz I put some specifics on my thoughts inline. Happy chat in voice on this, it's potentially a bit confusing (and open to opinion).

dtbuchholz:
@joewagner ah yes, that all makes sense. thanks for the explainers.

okay, i've reverted the version & dep bumps for local/node-helpers; only the SDK and CLI have been changed there.

joewagner (approved):
💪

PR #109, opened 2023-12-14
(sdk) update `package-lock.json` with sdk version

joewagner:
#108 was merged without running `npm install` hence the package-lock.json file has the wrong sdk version.  This updates package lock.


dtbuchholz (approved):
Dang, I always forget that step after updating the version. My b

PR #122, opened 2024-02-02
(local) enable forking public chains

joewagner:
fixes https://linear.app/tableland/issue/ENG-353/allow-forking-of-existing-network-for-local-testing


.github/workflows/test.yml
```diff
@@ -53,3 +53,5 @@ jobs:
 
       - name: Test/Coverage 🧪
         run: TEST_TIMEOUT_FACTOR=3 npm test
+        env:
+          INFURA_KEY: ${{ secrets.INFURA_KEY }}
```
joewagner, line 57: We need an infura api key added as an env var for the tests to pass.

joewagner: @carsonfarmer can you add this key to the test env?

packages/local/src/main.ts
```diff
@@ -121,19 +120,56 @@ class LocalTableland {
...
+    const shouldFork = !!config.forkUrl;
+    const forkChainId = this._getForkChainId(config);
+
+    const hardhatCommandArr = [
+      "hardhat",
+      "node",
+      "--port",
+      this.registryPort.toString(),
+    ];
+    // eslint-disable-next-line
```
joewagner, line 136: eslint wants us to use the form `const x: T = val;` instead of the `as` casting.  This doesn't work for the build ts checking because of the typing of `process.env`.

packages/local/src/validators.ts
```diff
@@ -34,13 +43,16 @@ class ValidatorPkg {
...
-  start(registryAddress?: string): void {
+  start(config: StartConfig): void {
     const binPath = getBinPath();
     if (binPath == null) {
       throw new Error(
         `cannot start with: arch ${process.arch}, platform ${process.platform}`
       );
     }
+    this.validatorDir = config.shouldFork
+      ? this.validatorForkDir
```
joewagner, line 54: A lot of the validator config file is mutated during start up, but the fork config file hold default values for homestead so we don't have to do as many mods.  It also acts as an example of what a fork config should look like compared to a from scratch config.  This can be helpful to anyone reading our source who's curious.

joewagner (commented):
This is RFR.  I tested the cli out manually and all of the env vars work the same as the ts/js class config options.

packages/local/registry/hardhat.config.ts
```diff
@@ -8,6 +8,49 @@ import {
...
+  },
+};
+
+if (process.env.FORK_CHAIN_ID && process.env.FORK_URL) {
+  const chainId = parseInt(process.env.FORK_CHAIN_ID, 10);
+  const chainInfo = helpers.getChainInfo(chainId);
+  const chainName = chainInfo.chainName;
+
+  if (chainName === "matic") {
+    networks.hardhat.chains = {
```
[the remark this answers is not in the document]
dtbuchholz: does mumbai also require custom logic here?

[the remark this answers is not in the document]
joewagner: I didn't look too much at setting this up for testnets.  In theory they would be "non standard chains" according to hardhat, so something like this would be required for mumbai too.
Do you think people want to fork testnets?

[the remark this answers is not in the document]
dtbuchholz: Yeah, DIMO was looking TJ fork Mumbai

joewagner:
@carsonfarmer The infura api doesn't seem to work.  Did you add the key, or the full url?

carsonfarmer:
> @carsonfarmer The infura api doesn't seem to work. Did you add the key, or the full url?

Just the key itself...

dtbuchholz (approved):
🚀 sick feature!

carsonfarmer (approved):

dtbuchholz:
oh, here's the related docs PR: https://github.com/tablelandnetwork/docs/pull/187

PR #129, opened 2024-02-22
(cli,sdk) remove support for all goerli chains

joewagner:


dtbuchholz (approved):

dtbuchholz:
@joewagner the final test just passed so should be g2g to merge/publish. (note: i gave it a :thumbsup: since technically, you created the PR.)

--

also, the tests take ~25 mins now. i wonder why? 

it looks like after merging the LT startup/shutdown cleanups, the tests take 2.5-3x longer and are causing them to flakily fail (i had to run them a few times in this PR). 

perhaps it's the additional LT forking tests? it doesn't look like those were in (1) but were added in (2):
1) an earlier PR merged to main, 9m 33s: https://github.com/tablelandnetwork/tableland-js/actions/runs/7835645804
2) the LT cleanups PR that happened a few weeks after (1), 25m 39s: https://github.com/tablelandnetwork/tableland-js/actions/runs/8084731453

joewagner:
> also, the tests take ~25 mins now. i wonder why?

I believe it's the forking tests.  We could potentially break them into their own action and not require them for PRs if that seems ok to everyone?

dtbuchholz:
Fine by me!

PR #135, opened 2024-04-11
(local) bump sdk dep and skip fork tests

joewagner:
The `local` package should get it's `sdk` dependency updated, but the chain forking tests keep timing out in the dependabot PR.  This PR manually bumps the sdk and skips the fork tests.  
My thinking here is that we can always run the fork tests locally during development and review if a change potentially affects that feature. Running them locally takes 15-20 minutes, but for CI it's nice if it takes an hour or more which is too much.


dtbuchholz (approved):
makes sense to me!

PR #136, opened 2024-04-17
(local) fix deviation from prettier rule in config

joewagner:
I've been noticing that `local` extends the validator config and does so without including a new line character at the end of the file.  This means that if you accidentally commit while the validator is running you potentially commit a violation of this repo's prettier.

This PR includes the EOL character in the generated config file.


joewagner:
Looking at this with fresh eyes and it seems better to add the line ending only if the original file has a line ending.  That way folks with different prettier rules don't have a problem.
I'll update asap and move back out of draft.

joewagner:
Ok, ci passing.  I tested it out by linking studio locally. This is rfr.

dtbuchholz (approved):

PR #137, opened 2024-04-18
Update to ethers v6

### LINKED ISSUE #77: [TABJS-3] SDK cannot be used with ethers.js v6
## context

A user reported this bug in Discord here [https://discord.com/channels/592843512312102924/1077277654132215960](https://discord.com/channels/592843512312102924/1077277654132215960)

## initial report

The initial report is from a user running the basic sdk example from the docs:

```
import { Database, helpers } from "@tableland/sdk"
import { Wallet } from "ethers"
import * as dotenv from "dotenv"
dotenv.config()

async function main() {
    const privateKey: any = process.env.PRIVATE_KEY
    const rpc: any = process.env.RPC_URL
    
    const wallet = new Wallet(privateKey);
    const provider: any = helpers.getDefaultProvider(rpc)
    const signer: any = wallet.connect(provider);    
    interface Schema {
        id: number;
        name: string;
    }
    
    const db = new Database<Schema>({ signer: signer })
    
    const prefix: string = "my_sdk_table";
    
    const { meta: create } = await db
        .prepare(`CREATE TABLE ${prefix} (id integer primary key, name text);`)
        .run();
    
    console.log(create.txn?.name)
}

main();
```

This results in: `Error: RUN_ERROR: signer.getChainId is not a function`

## details

Ethers.js v6 was recently released [https://docs.ethers.org/v6](https://docs.ethers.org/v6)
With this release the signer interface no longer has the `getChainId` method.  The chainId is potentially available elsewhere within the ethers v6 signer, but we might want to decide if there's a more robust way to get the chainId.
The v3 SDK had the requirement that a specific network had been connected to, which made this problem a little easier.

## ideas for a solution

* We could potentially try to sniff out what kind of interface the passed in signer supports.  The  downside I see here is that this could be cumbersome.
* We could determine the chainId from the query.  I *think* this makes sense, but I'd have to dive into implementation to be sure.

## additional comments

* This is blocked by hardhat supporting ethers v6.  I'm gonna move this to cycle 3, but I don't know how long it will take for hardhat to introduce support
* Looks like we will have to wait until https://github.com/ethers-io/ethers.js/issues/3734 is fixed to properly fix this and upgrade this package to Ethers v6
* Hardhat has beta support for ethers v6, but some of the evm-tableland open zeppelin dependencies have peer dependencies that require ethers v5.  So we will have to wait for something to change in OZ, as such I'm moving this out to cycle 6
* This has turned into a pretty difficult update because of all of the dependency requirements.  Moving this out another cycle.

<sub>From [SyncLinear.com](https://synclinear.com) | [TABJS-3](https://linear.app/tableland/issue/TABJS-3/sdk-cannot-be-used-with-ethersjs-v6)</sub>

dtbuchholz:
## Summary

- Migrates all code to ethers v6
- Adds Polygon Amoy, deprecates Mumbai

Closes #77.

## Details

- Primarily, refactoring for all code to use ethers v6, which required considerable testing to resolve nonce issues with tests. See the `getAccounts` method in the LT package for details—this **must** be used in all tests for ethers v6 to work with harhat when there are many txs in a row.
- Also, added Polygon Amoy support. Note this also changed the chain names for Polygon to use `polygon` vs. `matic`.
- Additional logic tweaks due to refactoring (will comment for awareness).

Note: node 16 has been dropped in hardhat & `@tableland/evm`, so expect that test to fail.

## How it was tested

Updated all tests across all clients.

## Checklist:


packages/cli/package.json
```diff
@@ -48,16 +48,15 @@
   },
   "license": "MIT AND Apache-2.0",
   "dependencies": {
-    "@ensdomains/ensjs": "3.0.0-alpha.64",
-    "@tableland/node-helpers": "^0.0.2",
-    "@tableland/sdk": "^6.0.0",
+    "@tableland/node-helpers": "^1.0.0",
+    "@tableland/sdk": "^7.0.0",
     "@tableland/sqlparser": "^1.3.0",
     "@tableland/studio-cli": "^0.1.0-pre.3",
```
dtbuchholz, line 54: note: optionally, we need the studio CLI to get bumped before we publish this CLI (will comment on why down below).

packages/cli/src/cli.ts
```diff
@@ -44,8 +44,6 @@ export interface GlobalOptions {
   providerUrl: string;
   baseUrl: string;
   verbose: boolean;
-  ensProviderUrl?: string;
```
dtbuchholz, line 47: ENS has been dropped from the CLI

packages/cli/src/commands/controller.ts
```diff
@@ -82,9 +82,10 @@ export const builder: CommandBuilder<Record<string, unknown>, Options> = (
           });
 
           const reg = new Registry({ signer });
-          const res = await reg.setController({ tableName: name, controller });
+          const tx = await reg.setController({ tableName: name, controller });
+          const res = await helpers.getContractReceipt(tx);
```
dtbuchholz, line 86: ethers v6 will return a transaction response with a `bigint` type. thus, the `JSON.stringify(...)` down below errors out (with `do not know how to serialized BigInt`). thus, we can either process the bigint -> string or number...but i chose to just use the method we already have.

packages/cli/src/commands/info.ts
```diff
@@ -23,46 +23,35 @@ export const builder: CommandBuilder<Record<string, unknown>, Options> = (
...
     logger.log(JSON.stringify(res));
-    /* c8 ignore next 7 */
+    /* c8 ignore next 13 */
   } catch (err: any) {
-    logger.error(
-      typeof err?.cause?.message === "string"
-        ? err?.cause?.message
-        : err.message
-    );
+    if (err.message.match(/table name has wrong format/)) {
```
dtbuchholz, line 45: changed the logic to use the sqlparser and validated table names. to keep the errors consistent, I'm catching and using the same one we defined before.

packages/cli/src/commands/studio.ts
```diff
@@ -6,6 +6,7 @@ import type { GlobalOptions } from "../cli.js";
 export const command = "studio <sub>";
 export const desc = "Tableland Studio CLI commands";
 
+// @ts-expect-error polygon amoy vs mumbai is `temporarily` causing a type error
```
dtbuchholz, line 9: here's why the studio CLI should be published before this CLI, and then _this_ CLI needs to update its deps before it gets published.

the studio CLI still has mumbai as a chain, so once it drops mumbai for amoy, then this TS directive is unneded

packages/cli/test/controller.test.ts
```diff
@@ -24,10 +23,8 @@ const defaultArgs = [
   "local-tableland",
 ];
 
-const accounts = getAccounts();
-const wallet = accounts[1];
-const provider = getDefaultProvider(TEST_PROVIDER_URL, { chainId: 31337 });
-const signer = wallet.connect(provider);
+const accounts = getAccounts(TEST_PROVIDER_URL);
```
dtbuchholz, line 26: when you review the LT `getAccounts` method, you'll see I've changed it to accept _either_ and LT instance, or a URL. this is required because we **must** use a provider with custom settings. 

plus, this makes it a bit easier as compared to the previous logic, which had to be that way because of our custom port setups during tests. now, the custom ports are passed to the method as part of the URL.

packages/local/package.json
```diff
@@ -17,7 +17,7 @@
     "ethereum"
   ],
   "engines": {
-    "node": ">=14.0.0"
+    "node": ">=18.0.0"
```
dtbuchholz, line 20: bumped to node 18 b/c the new hardhat requires LTS / works with >=18

packages/local/src/util.ts
```diff
@@ -504,25 +504,45 @@ export const getValidator = function (baseUrl?: string): Validator {
 /**
  * Get all of the connected accounts available for signing transactions.
  * Defaults to RPC URL `http://127.0.0.1:8545` if no instance is provided.
- * @param instance An instance of Local Tableland.
+ * @param instanceOrUrl An instance of Local Tableland or a chain RPC URL.
  * @returns An instance of a Tableland `Validator` with the correct `baseUrl`.
  */
-export const getAccounts = function (instance?: LocalTableland): Wallet[] {
+export const getAccounts = function (
```
dtbuchholz, line 510: here's the `getAccounts` logic change

packages/sdk/README.md
```diff
@@ -10,10 +10,13 @@
```
dtbuchholz, line 10: i'll update the README separately...i _don't think_ much, if anything, has changed, but I'll take care of that when i update the docs site

packages/sdk/src/helpers/config.ts
```diff
@@ -2,7 +2,7 @@ import { type WaitableTransactionReceipt } from "../registry/utils.js";
 import { type FetchConfig } from "../validator/client/index.js";
 import { type PollingController } from "./await.js";
 import { type ChainName, getBaseUrl } from "./chains.js";
-import { type Signer, type ExternalProvider, getSigner } from "./ethers.js";
```
dtbuchholz, line 5: much of this is the core of the ethersv6 changes. this old branch has a lot of the changes already mapped out, and i followed the majority of them: https://github.com/tablelandnetwork/js-tableland/tree/joe/bump-ethers

packages/sdk/src/helpers/ethers.ts
```diff
@@ -1,20 +1,83 @@
...
+  blockTime: number;
+  blockNumber: number;
+}
+
+/**
+ * Fetches the current gas fee data for a connected network.
+ * @param signer A signer instance.
+ * @returns Current gas fee information for the network.
+ */
+export async function getFeeData(signer: Signer): Promise<FeeData> {
```
dtbuchholz, line 57: we had issues with doing _anything_ in `@tableland/evm` without fetching live fee data. i've included the logic here to fetch data for amoy, and i figured doing it for polygon mainnet as well wouldn't hurt. if neither of those chains are provided, then it hits the generic ethers `getFeeData` method.

packages/sdk/src/helpers/ethers.ts
```diff
@@ -82,28 +139,49 @@ export interface MultiEventTransactionReceipt {
  *
  */
 export async function getContractReceipt(
-  tx: ContractTransaction
+  tx: ContractTransactionResponse
```
dtbuchholz, line 142: this is where i had to refactor a bit. a couple of notes:
- event/log parsing is more difficult in v6. perhaps there's an easier way.
- here, i created a new type (will explain below) that builds upon the types in the subscription logic where the index of a `tableId` is mapped to the event type.
- in `getContractReceipt`, i use those types when i parse the args to get the index of a `tableId`
- i also added support for setting a controller and transferring a table

packages/sdk/src/helpers/ethers.ts
```diff
@@ -124,15 +202,82 @@ export async function getSigner(external?: ExternalProvider): Promise<Signer> {
...
-  const web3Provider = new Web3Provider(provider);
-  return web3Provider.getSigner();
+  const browserProvider = new BrowserProvider(provider);
+  // Calling `browserProvider.getSigner` leads to a loop of annoying logging:
+  // `JsonRpcProvider failed to detect network and cannot start up; retry in 1s
+  // (perhaps the URL is wrong or the node is not started)`. Calling
+  // `_detectNetwork` here ensures this doesn't happen, but follow here for
+  // this issue getting resolved in the future:
+  // https://github.com/ethers-io/ethers.js/issues/4377#issuecomment-2023855491
+  await browserProvider._detectNetwork();
```
dtbuchholz, line 212: in case it ever comes up—this is a super annoying issue. see the link for details, but TL;DR it logs an error message every second until a connection is made

packages/sdk/src/helpers/ethers.ts
```diff
@@ -124,15 +202,82 @@ export async function getSigner(external?: ExternalProvider): Promise<Signer> {
...
+ * Create a signer with a private key, a provider URL, and a chain. Optionally,
+ * pass the chain name or ID to create a static network and reduce calls made by
+ * the provider (by not checking the chain ID).
+ * @param privateKey The private key of the signer.
+ * @param providerUrl The URL of the provider.
+ * @param chainNameOrId The chain name or ID.
+ * @param options Optional settings for the provider. See ethersjs
+ * {@link JsonRpcApiProviderOptions} for more information.
+ */
+export function createSigner({
```
dtbuchholz, line 249: this is a new method, not really used anywhere. i think it's useful because pretty much [every docs snippet](https://docs.tableland.xyz/quickstarts/sdk-quickstart#2-connect-to-a-signer) starts with creating a pk signer. this just abstracts it away a little bit.

packages/sdk/src/helpers/subscribe.ts
```diff
@@ -30,21 +31,74 @@ type ListenerMap = Record<
...
 >;
 
-type ContractEventTableIdMap = Record<
-  // the key is the event name in the Solidity contract
-  string,
+/**
+ * Possible events that can be emitted by the Registry as defined in the
+ * ITablelandTables interface.
+ */
+export type ContractEventNames =
```
dtbuchholz, line 38: here's where new types are defined for the event/log parsing i mentioned above. basically, I'm strongly typing the the registry's event names to silence some TS compiler complaints.

packages/sdk/src/helpers/subscribe.ts
```diff
@@ -30,21 +31,74 @@ type ListenerMap = Record<
...
+ * Mapping of only the subscribable contract events to the event names and the
+ * location of `tableId` in the event arguments, plus, emitter keyword.
  */
-const contractEvents: ContractEventTableIdMap = {
+type SubscribableEventTableIdMap = EventTableIdMap<SubscribableEventNames>;
+
+/**
+ * Mapping of all ITablelandTables contract events and `tableId` event index.
+ */
+export const contractEventsTableIdx: ContractEventTableIdMap = {
```
dtbuchholz, line 82: before, there was a single object `contractEvents` object (further below) that tracked events to their `tableId` and `emit` name. 

with `contractEventsTableIdx`, it basically does the same but:
- includes create table events
- doesn't use the `emit` keyword
i suppose i could maybe combine them. anyways, the `contractEventsTableIdx` is used in the `getContractReceipt` above.

packages/sdk/src/helpers/subscribe.ts
```diff
@@ -260,7 +313,8 @@ export class TableEventBus {
         });
       };
 
-      contract.on(key, listener);
+      const topicFilter = contract.filters[key as SubscribableEventNames];
```
dtbuchholz, line 316: some additional logic i had to change for filtering events/logs in ethers v6

packages/sdk/src/validator/client/fetcher.ts
```diff
@@ -262,7 +262,7 @@ function fetcher<Paths>() {
                 init: mergeRequestInit(defaultInit, init),
                 fetch,
               })
-          )) as CreateFetch<M, Paths[P][M]>,
+          )) as unknown as CreateFetch<M, Paths[P][M]>,
```
dtbuchholz, line 265: nre TS was complaining if i didnt cast and then recast. i couldn't figure out why. but, if i changed it to `CreateFetch` to have a string value for either of the generics, the error went away. not a big deal tho.

packages/sdk/test/aliases.test.ts
```diff
@@ -2,16 +2,14 @@
...
 import { Database } from "../src/index.js";
 import { TEST_TIMEOUT_FACTOR, TEST_PROVIDER_URL } from "./setup";
 
 describe("aliases", function () {
   this.timeout(TEST_TIMEOUT_FACTOR * 10000);
   // Note that we're using the second account here
-  const [, wallet] = getAccounts();
-  const provider = getDefaultProvider(TEST_PROVIDER_URL);
-  const signer = wallet.connect(provider);
+  const [, signer] = getAccounts(TEST_PROVIDER_URL);
```
dtbuchholz, line 12: just another callout to the new `getAccounts` that we should use everywhere for avoiding nonce issues

packages/sdk/test/subscribe.test.ts
```diff
@@ -126,13 +122,12 @@ describe("subscribe", function () {
...
           const eventTableId = eve[2].toString();
-          const txn = eve[3];
-
+          const log = eve[3].log;
           match(from, /^0x[0-9a-fA-F]+$/);
           match(to, /^0x[0-9a-fA-F]+$/);
           strictEqual(tableId, eventTableId);
-          strictEqual(txn.event, "TransferTable");
-          match(txn.transactionHash, /^0x[0-9a-fA-F]+$/);
+          strictEqual(log.fragment.name, "TransferTable");
```
dtbuchholz, line 129: nre another random ethers v6 change that could be useful to know, if you're ever parsing events

packages/sdk/test/subscribe.test.ts
```diff
@@ -243,8 +238,16 @@ describe("subscribe", function () {
 
       await eventBus.addListener(`${tableName}`);
       const listeners = eventBus.listeners[tableIdentifier];
-
-      strictEqual(eventBus.contracts["31337"] instanceof Contract, true);
+      // TODO: understand why checking `instanceof Contract` doesn't work
```
dtbuchholz, line 241: note: i had this issue in the `@tableland/evm` package as well. although a contract is typed as a `BaseContract` in typechain, the compiler doesn't recognize it as matching the ethers `Contract` interface

dtbuchholz:
@joewagner this is a hefty one! it includes both the ethers v6 and amoy changes.

it took _forever_ to figure out that flakey nonce issue, but things finally work. note that we're dropping node 16 support, so that test doesn't matter.

I'm in the process of updating the studio with these changes. ideally, we:
1. publish the `sdk`, `node-helpers`, and `local`
2. publish the `studio-cli`
3. bump the `studio-cli` in this repo's `cli` package.
4. publish the `cli` package

packages/sdk/src/validator/client/validator.ts
dtbuchholz, line 1: note: i updated the docs file locally and then ran the SDK `build:api` command off of that, so this should match in the future once the docs site gets updated

joewagner:
> @joewagner this is a hefty one! it includes both the ethers v6 and amoy changes.
> 
> it took _forever_ to figure out that flakey nonce issue, but things finally work. note that we're dropping node 16 support, so that test doesn't matter.

Nice work!  This has been a long time coming. 
Dropping node 16 makes sense to me, looks like you removed it from the test workflow matrix.  Can we also remove it from the review workflow matrix and add version 20. https://github.com/tablelandnetwork/tableland-js/blob/main/.github/workflows/review.yml#L18

> I'm in the process of updating the studio with these changes. ideally, we:
> 
> 1. publish the `sdk`, `node-helpers`, and `local`
> 2. publish the `studio-cli`
> 3. bump the `studio-cli` in this repo's `cli` package.
> 4. publish the `cli` package

I'll start working through reviewing all of this, and let's plan on releasing early next week.

joewagner:
@dtbuchholz Looks like you need to run prettier.  Potentially related to #136

dtbuchholz:
(@joewagner rebased this on the latest changes from main, looks like that fixed the review workflow issue!)

packages/cli/src/commands/studio.ts
joewagner: I think this makes sense. 
To be explicit, the plan is:
1. publish the sdk, node-helpers, and local
2. update sdk dep and publish studio-cli
3. update studio-cli dep and publish this cli

Does that sound right?  If so, let's publish a `pre-` version of the sdk, then open a pr for the studio-cli update, then update this PR with the new studio-cli version.

packages/cli/src/utils.ts
```diff
@@ -39,8 +44,8 @@ export function getLink(chain: helpers.ChainName, hash: string): string {
     }
     return `https://etherscan.io/tx/${hash}`;
   } else if (chain.includes("polygon")) {
-    if (chain.includes("mumbai")) {
-      return `https://mumbai.polygonscan.com/tx/${hash}`;
+    if (chain.includes("amoy")) {
+      return `https://www.oklink.com/amoy/tx/${hash}`;
```
joewagner, line 48: As of a few days ago https://amoy.polygonscan.com/ is up and running.  Does it makes sense to use that?

packages/cli/src/commands/studio.ts
dtbuchholz: @joewagner cool. so do i need to change the SDK version in this PR to `...-pre.0`? or, can we just publish the SDK "offline" as `pre.0` and then proceed with what's outlined above?

joewagner: Publishing pre versions "offline" is what I've done in the past.
I wrote a playbook in Notion, but it's basically `cd` into the `packages/sdk` directory, publish with npm (not lerna) and make sure to include `--tag next`

dtbuchholz: gotcha. but, my username doesn't have the ability to publish to npm, tho. maybe you could run that step for me?

dtbuchholz: oops, forgot joe's a bit OOO this week. @asutula or @carsonfarmer—if either of you have npm access, could you help publish a pre version of the SDK only? it'd be`7.0.0-pre.0`, and the process is described above.

(i need this to get the Studio changes going that use this PR, and i also want to get DIMO the new SDK w/ amoy so they can keep building.)

packages/sdk/src/helpers/ethers.ts
```diff
@@ -106,6 +106,23 @@ export function isPolygon(chainId: number | bigint): boolean {
...
+ * `eth_getMaxPriorityFeePerGas`, or `eth_estimateGas`, which are used by the
+ * built-in ethers `getFeeData` method.
+ * @param chainId The chainId of the network as a number or bigint.
+ * @returns A boolean that indicates if `getFeeData` can be used for the chain.
+ */
+export function chainCanGetFeeData(chainId: number | bigint): boolean {
+  const chainIdNumber = typeof chainId === "bigint" ? Number(chainId) : chainId;
+  // Hardhat, Optimism Sepolia, & Filecoin Calibration will return an RPC error
+  return !(
+    chainIdNumber === 31337 ||
```
joewagner, line 120: @dtb I'm wondering if we know for sure that chains with id `31337` always don't have these methods?

joewagner: For background, there was a user in discord using that chainId for a private chain and they were using that chain id.  Also there are some other local dev chains that use this id, e.g. https://chainlist.org/chain/31337

dtbuchholz: hm, yeah that's a good point; we can remove 31337. the main thing this method does is silence an error for `The method "eth_maxPriorityFeePerGas" does not exist / is not available.', data: {…}}`, which will get logged in the console of the Studio/web apps on those 3 chains. but, who cares if it gets logged with LT tbh.

some additional context: `anvil` and `hardhat` don't support the `eth_maxPriorityFeePerGas` method but do support `eth_gasPrice`. e.g., when you try to call the native ethers `getFeeData` on hardhat, the `eth_maxPriorityFeePerGas` method logs an error, but the tx still gets executed. aka we don't _need_ to include 31337 because it'll still work fine.

dtbuchholz: also, we _could_ adjust [this method](https://github.com/tablelandnetwork/tableland-js/blob/2c3b92be798309464b34fe8b37c58fccf2570c7e/packages/sdk/src/helpers/ethers.ts#L93) to destructure the `gasPrice` too. we just have to handle logic to avoid an error for `Cannot send both gasPrice and maxFeePerGas params`. e.g.:
```js
// also get the gas price and conditionally pass it
const { gasPrice, maxFeePerGas, maxPriorityFeePerGas } = await getFeeData(
  signer
);
if (maxFeePerGas != null && maxPriorityFeePerGas != null) {
  opts.maxFeePerGas = maxFeePerGas;
  opts.maxPriorityFeePerGas = maxPriorityFeePerGas;
} else if (gasPrice != null) {
  opts.gasPrice = gasPrice;
}
```
to get the gas price, we'd have to call `eth_gasPrice` directly to avoid the logging error from above.

joewagner: sticking to the removal of 31337 sounds good to me.  If it's annoying in LT we can stop logging those messages in LT? (If that makes sense, we do it in another PR)

dtbuchholz: works for me!

joewagner:
@dtbuchholz I just pushed a commit with some updated versions of the studio.  I think we are gtg for Monday.

The only notable thing is that since we have two monorepos with a circular dependency, I had to do something a little hacky to enable us to publish everything simultaneously next week.
I had to temporarily move the sdk and local out of the monorepo and publish a `next` version of those so that the studio could properly `npm install` and `npm run build`.  After that I published the `next` versions of studio without issue, which then allow us to `npm install` and `npm run build` this monorepo normally.
The only resulting weirdness is that I had to add two lint ignores and ts ignores to the local up.ts file.  This is because in the monorepo there's no error, but when the directory gets moved out of the npm workspace there is an error. 

tl;dr; we can either just leave them incase we do something like this again, or remove them once all the `latest` releases are published.

dtbuchholz:
gotcha. i suppose if leaving those files doesn't impact anything negatively, then why not do so...just in case we have a major bump in the future, which I'm guessing will happen

